PR #1, opened 2023-03-02
feat: add rpm-ostreed.conf

castrojo:
This belongs here now and will let us remove that inline sed from the containerfile


castrojo:
Need to rename a bunch of stuff so setting this to draft

Issue #17, opened 2023-03-15
Justfile missing in latest builds


castrojo:
Looks like the justfile isn't being copied over, need to investigate.

castrojo:
Ok it looks like we need to build the RPM here so it can go into main and then to the downstream derivatives?

bsherman:
Closed by #18 's merge.

Also, i tested both `silverblue-main:38` and `silverblue-nvidia:38` to verify not only is `just` present, but also the included `justfile` is what was expected per the respective images.

PR #18, opened 2023-03-24
feat: New akmods build process

dhoell:
See my PRs in -main, -akmods, -nvidia. This change breaks -nvidia, as it tries to install ublue-os-just again.

- Build ublue-os-just in -config
- Adjust build time to be more in line with quay.io/fedora-ostree-desktops

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,14 +6,14 @@ on:
     paths-ignore:
       - '**.md'
       - '**.txt'
-push:
+  push:
     branches:
       - main
     paths-ignore:
       - '**.md'
       - '**.txt'
   schedule:
-    - cron: '19 20 * * *'  # 7:20pm everyday    
+    - cron: '20 19 * * *'  # 7:20pm everyday    
 env:
     IMAGE_NAME: config
     IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}


.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ on:
       - '**.md'
       - '**.txt'
   schedule:
-    - cron: '20 19 * * *'  # 7:20pm everyday    
+    - cron: '0 5 * * *'  # 5 am everyday
```
bsherman, line 16: Eh I asked about the 7am in another place: https://github.com/ublue-os/main/pull/87#discussion_r1152663602

But same question as this would leave only 1 hour of buffer after the 4am UTC completion of builds.

bsherman:
Also a note, this would address #17  as this changeset now builds the `ublue-os-just` rpm.

.github/workflows/build.yml
bsherman: Question was answered in Discord and this seems good to me.

bsherman (approved):

castrojo (approved):

Eeyoresyndrome:
> Also a note, this would address #17 as this changeset now builds the `ublue-os-just` rpm.

I can confirm that on Kinoite 38-Main I have `just` now.
```Added:
  ublue-os-just-0.1-1.fc38.noarch```

PR #21, opened 2023-04-12
added some systemd services, power targets, udev rules and flatpak gui notifications

trytomakeyouprivate:
I am unsure what the flatpak system updates are supposed to mean. Are these flatpaks for all users?

I added a nice GUI notification displaying the amount of updated Apps and packages (only if greater than zero), I added udev rules activating battery and ac targets for use. 

I added another currently unused "full battery" target, that could be used in exchange for the "ac target", to ensure that the update will succeed.

A config for this, for PC users, could be useful too.

### DIFF
--- a/files/etc/polkit-1/rules.d/cups-passwordless.rules
+++ b/files/etc/polkit-1/rules.d/cups-passwordless.rules
@@ -1,0 +1,6 @@
+polkit.addRule(function(action, subject) {
+    if (action.id == "org.opensuse.cupspkhelper.mechanism.all-edit" &&
+        subject.isInGroup("wheel")) {
+        return polkit.Result.YES;
+    }
+});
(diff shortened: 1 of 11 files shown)


joshua-stone:
Hey, thanks for sending in this PR!

I think there's a lot going on here for one PR. Can you break this up into smaller, individual PRs that focus on one feature at a time?

trytomakeyouprivate:
Yes totally. I dont currently know how to do that... but I will try

trytomakeyouprivate:
Yes totally. I dont currently know how to do that... but I will try

@Linwer yes GNOME-Software and Discover also deal with firmware updates, rpm-ostree and flatpak updates. But I thought it was meant to disable these services to not hog resources?

Because if software is the center of usage (what the Fedora people want, also with KDE Discover) all these updating mechanisms are duplicates and not needed. But these Stores are very slow, resource-intensive and in my experience very inconsistent compared to the package managers themselves.

trytomakeyouprivate:
closed for now. I still havent found out how to setup git, 99% there but weird issues. Also the polkit rules have some syntax error so they break the plasma session. Maybe its the `""` around the polkit process.

trytomakeyouprivate:
comment: siosm plans on setting up at least automatic updates rpm-ostree and flatpak as a systemd service afaik. Automatic updates for sure.

So scripts for
- sudo waydroid upgrade
- firmware, dependend on dialog opt-in and full battery

are needed. Also a systemd service setup to detect restricted networks would help stop updates when you use your phone for example, working on that

PR #23, opened 2023-04-21
fix: update wooting rules

castrojo:


bsherman (commented):
This is a mostly good looking change. Adds a few devices, but also they change the group accessing these devices from `wheel` to `input`, which I generally like because then non-admin users can access these devices.

However.... by default, `input` group doesn't even exist in silverblue/ublue... and I looked at the `70-u2f.rules` which uses group `plugdev` which also doesn't exist.

We may need a story around creating these groups (`input` and `plugdev`) and adding users to them by default, if we intend users to use them.

atimeofday:
It should maybe be noted that this is not just adding or tweaking things, it is updating the rules to match the full current list directly provided by Wooting. Dunno if that changes anything about when/how to merge

castrojo:
Should we be adding extra users? Or would it make sense for me to modify the group to be wheel instead?

bsherman:
I think we should be trying to get users into an input or plugdev group rather as these rules are written rather than using wheel. It’s better because if someone sets up a non-admin user for gaming etc, they can still have access to the devices. On Apr 29, 2023, at 9:54 AM, Jorge O. Castro ***@***.***> wrote:﻿
Should we be adding extra users? Or would it make sense for me to modify the group to be wheel instead?

atimeofday:
If this may take a while to sort out due to important long term maintenance philosophy, could lines 59-70 be merged independently since they are for a new device doing its own thing? I had not realized this would not be a trivial addition to fix an entirely unsupported device, and I would like to be able to configure my keyboard. Does the same concern apply to the new device by adding / messing with a new group?

bsherman:
I was wrong that `input` group is missing in Silverblue... but I'm not wrong that there's something weird about it.

Thus I'm trying to understand the situation with groups on Silverblue... on my current ublue system:

1) `systemd-sysusers` is the process that creates system users/groups, this should usually happen on first boot? maybe it updates if there's additions to `/usr/lib/sysusers.d` 

2) rpm-ostree has the "managed" groups in `/usr/etc/group` (eg, i think this get created/appended when we install packages in our Containerfile builds) ... rpm-ostree apparently merges this with `/etc/group` somehow (just read this, not sure what code makes it happen)

3) there also exists `/usr/lib/group` which is where, as I care about with respect to the PR above, `input` group resides (not in `/etc/group` or `/usr/etc/group`)

4) `getent group` will show the full, merged set of groups in a system... and `genteng group | grep input` shows for certain, the system can resolve `input:x:104:`  from `/usr/lib/group` so... we can at least read that...

5) but here's the crazy part... `sudo usermod -a -G input MYUSERNAME` succeeds, or at least says it does (using an obviously non-existing group fails with group does not exist)... BUT..... even after rebooting, my MYUSERNAME is NOT a member of the `input` group... 

So... it seems like shadow utils is not working right possibly? it should probably create a masking `input:x:104:MYSUSERNAME` line in `/etc/group` when the above `usermod` command runs... since clearly `/usr/` is read-only... 

this is where i think its time to test on pure silverblue. 

And... I intend to break out the group changes from this PR so that this PR will just contain the added device rules, not a change to groups... we should solve group mysteries in another issue.

bsherman:
Changed my mind... I've submitted a simplified PR ( #30 ) with just the new wooting rules...

We can keep this PR until the groups concern is resolved.

Arcitec:
Let's enumerate our options.

1. If we add all users to `input` group as originally planned, it's the exact same as opening device access to everybody. So there's no security difference.

2. Opening device access to all users means we don't need to add users to a silly `input` group. So that's way cleaner. It's how Valve solved it.

3. Even if adding `all users` to read/write privilege for the keyboard devices, are there really any security implications? The user is already typing at the keyboard. They know what they're typing. And malware has more important things to do than hunting for wooting users to read the device to keylog all the cool wootbros in the world.

4. The last option is to just not do anything. Users who aren't in `input` will have to add themselves to it, and take full responsibility for that security.

For that reason, I think it makes sense to do option 2. Because option 1 sure as hell doesn't make sense since it's an identical outcome with MORE work for us.

bsherman:
STOP THE PRESSES! 

THE MYSTERY HAS BEEN SOLVED!


TL;DR
1. @KyleGospo was the original contributor of `80-wooting.rules`  ( see https://github.com/ublue-os/udev-rules/commit/eef7b5ae1fe46e8c5c5f376866aa0b83f3edda00  )
2. He explicitly chose to use `wheel` instead of `input` (which Wooting does by default) because
    1. `input` is not default on Fedora
    2. `wheel` meant only admins could reconfigure the keyboard
3. this means we have no "bug" to "fix"

My proposal: Add a comment explaining the rationale here and call it done. Future PRs for wooting should be simple.


Full story ... go read a lot of discord chat...

https://discord.com/channels/1072614816579063828/1072617059265032342/1107549060237361192

PR #25, opened 2023-04-24
Created just some udev rules and fitting systemd targets

trytomakeyouprivate:
What I did: create udev rules detecting 
- battery full and charging
- general AC attached or "on battery"
- battery low

They trigger systemd targets that can in further steps be used from the user to add powersaving or -unlocking scripts, disabling CPU cores, throttling clockspeed and so on. It should run fully automatic, which is really nessecary.

If there are syntax errors please correct them ;D

### DIFF
--- a/files/etc/sysctl.d/90-overrides.conf
+++ b/files/etc/sysctl.d/90-overrides.conf
@@ -1,0 +1,1 @@
+vm.max_map_count=2147483642
--- a/files/etc/udev/rules.d/90-ac-states.rules
+++ b/files/etc/udev/rules.d/90-ac-states.rules
@@ -1,0 +1,5 @@
+# This module detects if AC is attached or not and activates systemd targets
+# The systemd targetscan then be used for things like batterysaving services
+
+SUBSYSTEM=="power_supply", KERNEL=="AC", ATTR{online}=="0", RUN+="/usr/sbin/systemctl start battery.target"
+SUBSYSTEM=="power_supply", KERNEL=="AC", ATTR{online}=="1", RUN+="/usr/sbin/systemctl start ac.target"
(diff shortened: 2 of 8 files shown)


trytomakeyouprivate:
In the next step I would create systemd services for 

- fwupd (depending on battery-full.target)
- batterysaver (depending on battery.target)
- batterysaver-low (depending on battery-low.target)
- batterysaver-undo (depending on ac or battery-full, discussable)

I think these should be in /etc to be user-configurable.

In general they apply to all DEs. Is there a difference what systemd services have to be enabled or not? They should not cause any unwanted problems, for example on Desktops, but you never know. They should be possible to disable and enable.

Also The services should trigger scripts the user is supposed to setup themselves, as measurements will vary a lot. We could gather some presets for different CPUs or GPUs for example.

TLP broke my ports on my new laptop, so this is not the simplest and best solution. Power-profiles-daemon (integrated in KDE) also does not work on my AMD laptop.

But there are many ways to improve batterylife. If TLP works for people they can use that, they can also disable cores, throttle speeds and more.

akdev1l:
My first reaction is that automatic firmware upgrades sound unnecessary and dangerous. That seems like something that should require user consent/interaction as it can introduce bugs to devices that may have already been working fine. 

That part I would be against merging as it can be problematic. 

The battery stuff seems ok but too opinionated to be part of the upstream main image imo. I wouldn't want any of this stuff on my desktop system that doesn't even have a battery. 

Having a custom image with all these changes applied seems fair game but I think that's better left off as a downstream consumer of the uBlue images

trytomakeyouprivate:
Yes I understand that. But as far as I know these services have to be enabled anyways, right?

Systemd services in `/usr/lib/systemd` are not enabled by default. They need to, by symlinking them to the `/etc/systemd/` directory.

So yes having that may be annoying on Desktops, but

1. Its not enabled
2. It runs an empty script users have to setup themselves. 

Where to put that script is discussable, I would say a root directory like /etc/ somewhere. And we could add many presets making power optimisation very easy.

The majority of users (uploading data to linux-hardware.org) uses portable devices for Fedora

image

PR #42, opened 2023-05-13
refactor: use standard path for just plus...

### LINKED ISSUE #41: standardize path and behavior of ublue-os justfile
@bsherman @EinoHR After some discussion on Discord, a good pattern emerged to avoid forks overwriting each other's justfiles randomly: Every image instead has its own separate justfile, which the user can include/merge a-la-carte.

Currently we have these:

```sh
touch ~/.justfile
cat /usr/share/ublue-os/just/main.just >> ~/.justfile
cat /usr/share/ublue-os/just/custom.just >> ~/.justfile
```

The `main` is meant to be the main config, which currently comes from the `config` repo. Its purpose is to contain the "core" commands.

The `custom` is from the startingpoint template repo, which is where the repo maker adds their own. It includes absolutely *nothing* from `main` and therefore never risks going out of sync with upstream `ublue-os/config`.

We'll need 4 minor changes in how the main justfile is built:

- Target location: `/usr/share/ublue-os/just/main.justfile`
- Script update: `ublue-os-just.sh` needs to use the new justfile path. The purpose of that script is currently to auto-install the justfile on login if the user (even root) doesn't have a `~/.justfile`, and it does nothing if the user has the file. So its behavior is already fine. It only needs a tweak to the path. Well, it also uses illegal non-POSIX shell syntax (only works in bash and zsh), so I'll propose a fixed version below.
- The script currently places itself in `/etc/profile.d` on the image which is incorrect. It needs to write itself to `/usr/etc/profile.d`, and from there "systemd" will take care of syncing it into the user's real `/etc` folder.
- We may want to make it merge every OS-level justfile automatically, not just the main one. But that's tricky with a `profile.d` script since we can't know which shell we're running under and therefore can't use bashisms for example... so leaving that off for now.

Alright here's the fixed `ublue-os-just.sh`:

```sh
# Add uBlue's justfiles to user if they don't already have a justfile.
if [ ! -f "${HOME}/.justfile" ] && [ -f /usr/share/ublue-os/just/main.just ]; then
  cp /usr/share/ublue-os/just/main.justfile "${HOME}/.justfile"
fi
```

In time I will probably contribute to the other repos too, but at the moment I am still busy and uncomfortable with the structure of main, config and nvidia. I'm starting to study them though! :)

_Originally posted by @Arcitec in https://github.com/ublue-os/base/discussions/149#discussioncomment-5889873_

bsherman:
This changes the ublue-os-just RPM by:
- using /usr/share/ublue-os/just path (no double ublue-os)
- renames "justfile" to "main.just"
- adds "nvidia.just" and "custom.just"
- improves profile script to smartly add appropriate just files

Closes #41


Arcitec:
Woo! This is looking really awesome! :)

1. Should we include `nvidia.just` here like? Doesn't the addition of that file belong to our NVIDIA repo? Although I suppose it's easier to manage it here! And you did a great job with the nvidia rpm check to make sure only NVIDIA users get it! :)
2. The `custom.just` is added by `startingpoint` during layering, so I was surprised to see a blank one being included in the RPM here. But it actually seems kind of nice to put that blank dummy file here, to "enforce that rule". It's a harmless empty file if people don't use `startingpoint` and it can give people ideas if they build their own containers manually... so I like it. :)
3. The patching method of the `profile.d` script needs some fixes for non-bash shells (for example `zsh` cannot append to non-existent files which is why we must `touch` first, and we can't trust that `$HOME` has a value which would lead the current script to attempt a write to `/.justfile`), and we could also avoid creating those permanent global variables when they aren't needed. Posting below. Feel free to change anything below! :)

```sh
# Only process users with home directories, who are lacking a justfile.
if [ ! -z "$HOME" ] && [ -d "$HOME" ] && [ ! -f "${HOME}/.justfile" ]; then
  UBLUE_JUST=/usr/share/ublue-os/just
  USER_JUSTFILE="${HOME}/.justfile"
  touch "${USER_JUSTFILE}"
  if [ -f ${UBLUE_JUST}/main.just ]; then
    cat ${UBLUE_JUST}/main.justfile >> "${USER_JUSTFILE}"
  fi
  if [ -f ${UBLUE_JUST}/nvidia.just ] && [ rpm -q xorg-x11-drv-nvidia ]; then
    cat ${UBLUE_JUST}/nvidia.just >> "${USER_JUSTFILE}"
  fi
  if [ -f ${UBLUE_JUST}/custom.just ]; then
    cat ${UBLUE_JUST}/custom.just >> "${USER_JUSTFILE}"
  fi
fi
```

This would work on every shell. :)

bsherman:
this makes sense except for this line:

```
cat ${UBLUE_JUST}/main.justfile >> "${USER_JUSTFILE}"
```

I was originally using `cp` which should work in bash or zsh and should eliminate need for touch, unless you are trying to cover the edge case where main.just doesn't exist, which really can't happen unless users remove the main.just file but keep using the profile script.

Arcitec:
> unless you are trying to cover the edge case where main.just doesn't exist

Yep, it was just for the paranoid scenario where someone makes their own image without `main.just` existing anymore, maybe they add a `rm` command in their own scripts to delete it for example.

bsherman:
> > unless you are trying to cover the edge case where main.just doesn't exist
> 
> Yep, it was just for the paranoid scenario where someone makes their own image without `main.just` existing anymore, maybe they add a `rm` command in their own scripts to delete it for example.

Cool. I'm content with being paranoid.

build/ublue-os-just/ublue-os-just.sh
```diff
@@ -1,10 +1,11 @@
...
 
-UBLUE_JUST=/usr/share/ublue-os/just
-USER_JUSTFILE="${HOME}/.justfile"
-if [ ! -f "${USER_JUSTFILE}" ]; then
+if [ ! -z "$HOME" ] && [ -d "$HOME" ] && [ ! -f "${HOME}/.justfile" ]; then
+  UBLUE_JUST=/usr/share/ublue-os/just
+  USER_JUSTFILE="${HOME}/.justfile"
+  touch "${USER_JUSTFILE}"
   if [ -f ${UBLUE_JUST}/main.just ]; then
-    cp ${UBLUE_JUST}/main.justfile "${USER_JUSTFILE}"
```
Arcitec: Nice. It's easy to get blind from these filenames since we're writing `.just` files to `.justfile`, so I didn't notice that either!

Arcitec (approved):

bsherman:
nvidia's related PR is prepped: https://github.com/ublue-os/nvidia/pull/101

build/ublue-os-just/ublue-os-just.sh
```diff
@@ -7,7 +7,7 @@ if [ ! -z "$HOME" ] && [ -d "$HOME" ] && [ ! -f "${HOME}/.justfile" ]; then
   if [ -f ${UBLUE_JUST}/main.just ]; then
     cat ${UBLUE_JUST}/main.just >> "${USER_JUSTFILE}"
   fi
-  if [ -f ${UBLUE_JUST}/nvidia.just ] && [ rpm -q xorg-x11-drv-nvidia ]; then
+  if [ -f ${UBLUE_JUST}/nvidia.just ] && [ rpm -q ublue-os-nvidia-addons ]; then
```
Arcitec, line 10: That's a really good idea, and good reason for it!

Arcitec (approved):

castrojo (approved):

PR #44, opened 2023-05-15
fix: correct mistaken install of empty file to /etc/profile.d

bsherman:


Arcitec (approved):
Thank you! I read through and all source identifiers are in order now. Excellent. :)

PR #45, opened 2023-05-15
fix: update wooting rules, use modern security standards

Arcitec:
Wooting's current rules are a bit dangerous. I have fixed the rules [and reported the issue upstream](https://github.com/WootingKb/wootility-issues/issues/114#issuecomment-1548009293). I've also created a small development tool we use to instantly generate fixed rules until upstream has fixed the issue.

Anyone who adds themselves to the `input` group is currently able to keylog the Wooting keyboard.

This is what the old/upstream Wooting rules allow:

- User A is sitting at the keyboard.
- User B is logged in via SSH and is in the `input` (or `wheel`) group.
- User B is keylogging everything User A is typing.

We must add the `uaccess` tag, which makes systemd automatically grant device access to **only** the _currently_ logged in, _local_ (non SSH) user, to prevent broader keylogging.

It's done by removing your current `MODE` and `GROUP` and replacing them with this:

```
, MODE="0660", TAG+="uaccess"
```

Every file in `/dev/input` is owned by `root:input` and the `MODE` therefore checks that the user is either the owner or part of the same group of the file (`input` group), and the `TAG` ensures that only the physical user has access.

**Important:** When switching to the uaccess method, it's important that the rules are in a file that's named something numerically before `73-seat-late.rules` by the way! So I took care of that too.

https://github.com/systemd/systemd/issues/4288#issuecomment-348166161

The unofficial standard technique is to use the number `70`. So the rules should be in a file named `70-wooting.rules`.

These "best security practices" are also documented here: https://wiki.archlinux.org/title/Udev#Allowing_regular_users_to_use_devices


Arcitec:
This is ready to ship as soon as @castrojo gives the okay.

PR #58, opened 2023-05-28
feat: support new akmod signing key in just files

bsherman:
This add secure boot signing key for akmods to the `main.just` and renames the legacy version of the command in `nvidia.just` to be specific for older nvidia-only key.


bsherman:
Keeping this  draft until akmods stuff is building and ready to go.

bsherman:
This should be merged before https://github.com/ublue-os/main/pull/234

castrojo (approved):

Issue #46, opened 2023-05-19
Add android udev rules


gerblesh:
### Describe the package

I'd like to request adding the android udev rules because one has to go out of their way to install them manually on fedora, and on an immutable system it's near impossible without forking and creating your own image. The android udev rules allow developers to interact with their android devices and people to install alternative operating systems on their phone, as well as use diagnostic tools for android devices. Adding these to the base image will cause significantly less headaches for those working with android devices using uBlue.

### Information on the package

https://github.com/M0Rf30/android-udev-rules is the upstream repo for the rules, installation is also detailed in the repo.
I don't really know how i'd make a merge for this as I don't currently understand how the other custom udev rules for game controllers and such work, maybe someone can help me understand this better.
Thanks for your time!

### Image

All Images

gerblesh:
ok, so AFAIK fastboot, etc works by default without messing with users and groups. It's only ADB that needs those setup, I have a simple script in my own downstream image that sets this stuff up but I think it would be good to just leave it at the udev rules. Basically this means that most diagnostic tools + installing alternative operating systems on an android phone work by default, but adb (android debug bridge) which is mainly used for development of android apps and debugging, is not available by default because of the users/groups not being configured. This can easily be fixed by a couple of commands though.

castrojo:
Hey @Arcitec do you have cycles to look at this? I'm short on time and we also need to document how to submit these better. It'd be nice if we had a section in the README that any linux user could understand on making a PR it would be easier, it's still unclear to me what things I need to do to add rules in here. Thanks!

fiftydinar:
I believe that android-tools should be added as an package as default as well for complementing this package.

trytomakeyouprivate:
For me it was similar. I could also not use adb through distrobox, as the podman containers needed usb permissions and that was too much of a mess. I will try it though, and upload a guide.

In my experience I needed to run ADB as sudo, which is unnessecary with the correct rules. The same for LUKS decrypting and mounting or opening virt-manager.

PR #60, opened 2023-05-30
feat: add android udev rules to /etc/udev/rules.d

gerblesh:
partially solves: https://github.com/ublue-os/config/issues/46
Don't know how to update the RPM spec file yet, maybe i'll read up on it later


bsherman:
Copying thoughts here which I shared in chat...

I love the idea of this contribution... but I was looking at the upstream of  this file, https://github.com/M0Rf30/android-udev-rules , and it provides an RPM spec file.

It would be preferable to build that RPM (in COPR?) and install it directly in `main`.

PR #61, opened 2023-05-31
feat: create default justfile for users

bsherman:
When a user does not have a $HOME/.justfile, provide a default which uses `just includes` from our provided files as documented on the website.

https://ublue.it/guide/just/


EinoHR (approved):
Great!

castrojo (approved):

PR #62, opened 2023-06-04
Create udev rules for steering wheels

odintdh:
These rules are from [oversteer](https://github.com/berarma/oversteer/) and they set permissions to allow applications like Oversteer or PyLinuxWheel to configure Fanatec, Logitech, and Thrustmaster steering wheels.

These programs configure some attributes like rotation range and force feedback, exposed by the kernel driver through sysfs interfaces. Without these rules, Oversteer needs root permissions to configure these values.

### DIFF
--- a/files/etc/udev/rules.d/99-fanatec-wheel-perms.rules
+++ b/files/etc/udev/rules.d/99-fanatec-wheel-perms.rules
@@ -1,0 +1,28 @@
+# Match kernel name of device, rather than ATTRS{idProduct} and ATTRS{idVendor}
+# so we can access the range file and leds directory. Set rw access to these 
+# files for everyone.
+SUBSYSTEMS=="hid", DRIVERS=="ftec_csl_elite", GOTO="fanatec_device"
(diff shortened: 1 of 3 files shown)


castrojo:
Oh awesome, nice catch! Think this will need some RPM spec adjustments so let's see if someone can help us out with that.

bsherman:
> Oh awesome, nice catch! Think this will need some RPM spec adjustments so let's see if someone can help us out with that.

Yeah, would be good to get these added. However, I'd like to see it done by downloading from the [oversteer](https://github.com/berarma/oversteer/) repo, extracting the udev rules, and then placing them in an `oversteer` subdir within the ublue-os path as a sibling to `game-devices-udev` probably including the oversteer README there too.

Just good to ensure we always have latest updates from their repo and provide attribution.

bsherman:
Closing this as oversteer was packaged into COPR and added almost a week ago.

https://github.com/ublue-os/main/pull/366

PR #64, opened 2023-06-18
feat: add dnkmmr nix installer to justfile

EinoHR:
Add's @dnkmmr69420's Nix installer to the justfile with some warning attached.

https://github.com/ublue-os/startingpoint/issues/95

### DIFF
--- a/build/ublue-os-just/main.just
+++ b/build/ublue-os-just/main.just
@@ -34,3 +34,19 @@ update:
 fixscreenshare:
   cp /usr/share/applications/org.kde.xwaylandvideobridge.desktop $HOME/.config/autostart/
   
+nix-me-up:
+  #!/bin/bash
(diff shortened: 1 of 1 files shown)


EinoHR:
A Nix uninstaller / uninstall guide should also probably exist.

We have the uninstaller that was recently merged in startingpoint: https://github.com/ublue-os/startingpoint/pull/93, and dnkmmrs uninstaller: https://github.com/dnkmmr69420/nix-installer-scripts/blob/main/uninstaller-scripts/silverblue-nix-uninstaller.sh. I don't know which one we should use. The one in startingpoint does _look_ better but I'd be inclined to use dnkmmrs as that is paired with the installer.

EinoHR:
@castrojo opposes nix in this repo and dynamically pulling in the installer, closing.

Issue #63, opened 2023-06-17
Centralised Update System


IlChitarrista:
# Objective

The Objective of this suggestion is to reflect and develop a System that can Automatically Manage System Updates from a Centralised Environment without interfering with the usage of the OS and while providing some level of customisability and error reporting.

# Software Support

- rpm-ostree
- flatpak
- distrobox | (toolbox)
- (Fleek / Nix)
- (More?)

The fundamental systems that we need to support are rpm-ostree for the automatic updates while OS Upgrades (38 --> 39) are fine to be handled from Gnome Software as they normally should require more intentionality. Then Flatpak as that is the main method of acquiring software on Immutable systems. Distrobox, while often used for Development, can be a powerful tool to install software which might not have an OS Package or Flatpak (It should also be theoretically preferred to rpm-ostree)(This would be the case, for example, of Reaper).

## Extra Software?

We could also consider supporting Fleek / Nix but I would need someone to contribute extra info on that as I've never used such systems. I also highly incourage anyone to contribute some other system which would need support.

# GUI?

It would be great to have an OS which can be fully interacted both from a GUI and a CLI and this system could have some integration with Gnome Software although this would require a significant ammount of knowledge. This should also be a discussion upstream as pretty much no OS other than [Vanilla OS](https://handbook.vanillaos.org/2022/12/10/updates.html) implements this almost perfectly.

# Basic Flow of Execution

Single Check
```mermaid
graph TB
    A[Start] --> B{Configurable Timer}
    B -->|Every 24h| C[Check for Non-Interruption]
    B -->|Every Week| C
    B -->|Never| D[End]
    C -->|All Conditions OK| E[Check for System Updates]
    C -->|Any Condition Fails| D
    E -->|Updates Available| F[Handle Updates Linearly]
    E -->|No Updates| D
    F -->|Error| G[Notify User and Continue]
    F -->|No Error| H[Proceed to Next System]
    G --> H
    H -->|All Systems Updated| D
    H -->|More Systems to Update| F
```

Multiple Check
```mermaid
graph TB
    A[Start] --> B{Configurable Timer}
    B -->|Every 24h| C[Check for Non-Interruption]
    B -->|Every Week| C
    B -->|Never| D[End]
    C -->|All Conditions OK| E[Check for System Updates]
    C -->|Any Condition Fails| D
    E -->|Updates Available| F[Handle Updates Linearly]
    E -->|No Updates| D
    F --> G{Check for Non-Interruption\nSkip the First time}
[...]

PR #68, opened 2023-06-25
feat: added new update system

gerblesh:
Discussion: https://github.com/orgs/ublue-os/discussions/187
Solves: https://github.com/ublue-os/config/issues/63

after quite a bit of troubleshooting/fixing i've finally gotten the ublue-updater to build, and made a basic draft pull request to add the new update system.
This is by no means meant to be merged right away, and will likely need more changes/review (particularly moving the repo into the uBlue org)

Currently in the PR:
remove the old update services and update services RPM
pull in new update system from https://github.com/gerblesh/ublue-updater


gerblesh:
Update: basically everything works that needs to work as an MVP, updating, notification, logging failed updates, etc. It's already quite a bit more polished than the current update system we have in place. I'd say we just move the repo into the uBlue org and merge, that being said, I think it would be good to get some acknowledgement from image maintainers to make sure they are on board with this before merging

marcoceppi (changes requested):
Going to put a pause on this after the discussions in discord. The trend is that this is something better suited for bling and not as the default for main and those "upstream of custom" images.

gerblesh:
Closing this as it's out of scope for the uBlue main images and config, will make a PR in https://github.com/ublue-os/bling instead

PR #74, opened 2023-07-16
feat(just): add clean shortcut

castrojo:
We can add multiple things in here, but starting with a podman system prune.


akdev1l (approved):

bubuntux:
@castrojo  would it be better if it included the "-a" argument to remove all unused data?

castrojo:
Good idea, I added that in a follow up, thanks!

PR #80, opened 2023-07-19
Enable signed images

castrojo:


files/usr/etc/pki/ublue-os.pub
EyeCantCU, line 1: File path should be files/usr/etc/pki/containers/ublue-os.pub

EyeCantCU (approved):

PR #81, opened 2023-07-22
feat: add repair and uninstall unused flatpak maintenance tasks

### LINKED ISSUE #14: Investigate flatpak update maintenance
Not sure if we need to do an automated `flatpak repair` and or remove unusued, and update appstream, etc. We should make sure we're making it so users don't need to "remember" to do any kind of maintenance.

bsherman:
Improves upon the existing flatpak update service unit.

closes #14


castrojo:
Sorry I missed this but it's been a bit so just double checking that this is ready to go?

EyeCantCU (approved):
LGTM!

PR #84, opened 2023-07-23
feat: standardize flatpak/rpm-ostree update timers

### LINKED ISSUE #26: Do not use OnBootSec in the .timer file for updates since OnCalendar is already defined
The next run for the OnCalendar timer is calculated when the timer is first installed and is unaffected by previous runs triggered by the OnBootSec timer.

With the current implementation, if I reboot my computer at 05:50, the update service will be executed at 05:55 triggered by by the OnBootSec timer. It will then be executed again at 06:00, triggered by the OnCalendar timer. The previous run that was triggered by bootup will not affect the next run that happens at 06:00.

If your intent was to introduce a delay before the service is started, a better approach would be to use `ExecStartPre=/usr/bin/sleep 300` in the .service file.

If this is the intended behaviour, let me know and I will close the issue.

bsherman:
This adds an override for the Fedora stock rpm-ostreed-automatic.timer and sets it and both the user and system flatpak timers to always run at 4:00AM localtime. If the host machine is suspended or powered down   during that time, the timers will execute on next boot. 

The flatpak update services have been simplified to not request install and to be simple services like rpm-ostreed-automatic.service.

Closes #26


un1c3:
Looks good to me.

castrojo (approved):

PR #86, opened 2023-07-27
feat(just): add flatpaks to path

Renner0E:
The shebang is there because it would throw a [whitespace error](https://just.systems/man/en/chapter_47.html) without it.

### DIFF
--- a/build/ublue-os-just/main.just
+++ b/build/ublue-os-just/main.just
@@ -42,4 +42,18 @@ update:
 
 fixscreenshare:
   cp /usr/share/applications/org.kde.xwaylandvideobridge.desktop $HOME/.config/autostart/
-  
+
+add-flatpaks-to-path:
+  #!/bin/sh
+  echo 'Adding Flatpaks to $PATH'
+  sudo bash -c 'cat > /etc/profile.d/flatpak-bindir.sh' <<-'EOF'
+  if [ -n "$XDG_DATA_HOME" ] && [ -d "$XDG_DATA_HOME/flatpak/exports/bin" ]; then
+    pathmunge "$XDG_DATA_HOME/flatpak/exports/bin"
+  elif [ -n "$HOME" ] && [ -d "$HOME/.local/share/flatpak/exports/bin" ]; then
+    pathmunge "$HOME/.local/share/flatpak/exports/bin"
+  fi
+  if [ -d /var/lib/flatpak/exports/bin ]; then
+    pathmunge /var/lib/flatpak/exports/bin
+  fi
+  EOF
+  echo 'Restart your Terminal for changes to take effect'


akdev1l (commented):
Does this affect non-interactive shells? It probably shouldn’t

bsherman (changes requested):
I tested this locally to verify assumptions, and I'm uncertain what value this adds. I'd like to understand better before allowing approval.

@Renner0E can you clarify the purpose? I understand that flatpak/exports/bin is added to PATH... and yes, it works.

1) if approving I'd suggest changing the order of priority such that the user's personal flatpaks have higher PATH priority than system-wide

2) with this path exported... what changes this:
instead of running `$ flatpak run org.gnome.Contacts` the user can directly run `$ org.gnome.Contacts` 

I'm not if that's truly desirable at the "main" Universal Blue image level as it creates a non-standard way to run flatpaks, yet it's not doing what I usually hear people asking about (example: instead of running `flatpak run org.gnome.Contacts` wanting to run `gnome-contacts`).

Renner0E:
> @Renner0E can you clarify the purpose? I understand that flatpak/exports/bin is added to PATH... and yes, it works

It just makes it a little less annoying to launch flatpaks in the terminal. Nothing special it's just nice to have from my point of view.

> instead of running `flatpak run org.gnome.Contacts` wanting to run `gnome-contacts`).

I'd love this but it's probably not going to happen anytime soon because of [security concerns](https://github.com/flatpak/flatpak/issues/994#issuecomment-328154457)


> 1. if approving I'd suggest changing the order of priority such that the user's personal flatpaks have higher PATH priority than system-wide

Thanks for pointing that out! I Fixed it in a new commit.

AFAIK The first file/directory in your $PATH has the most priority. So you'd have to add it later because it gets inserted at the beginning of the line. Works on my machine™.

@akdev1l If I understood this correctly then this would also affect non-interactive shells aka. launching a script/subshell.
Arch Linux ships [this script](https://gitlab.archlinux.org/archlinux/packaging/packages/flatpak/-/blob/main/flatpak-bindir.sh) by default with the flatpak package. They probably know what they are doing and they know better than me.
I actually thought that it is normal to have flatpaks in the $PATH on linux because of this. Until I tried Fedora at least.

bsherman:
> Arch Linux ships [this script](https://gitlab.archlinux.org/archlinux/packaging/packages/flatpak/-/blob/main/flatpak-bindir.sh) by default with the flatpak package.

Ah, that clears up why this might be expected; I appreciate the context.


I feel like this is a nice-to-have change, but is more about making Universal Blue "Arch-like" than fixing something wrong or missing in Fedora. When faced with "maybe" situations, I always try to go back to https://universal-blue.org/scope/ and see if the change obviously fits with staying "close(ish) to Fedora" or not.

My vote is no, because I try to keep the top level Universal Blue "main/nvidia" images as close to stock Fedora as we can within our scope. However, this could be a nice inclusion for downstream images. Both Bazzite and Bluefin are doing LOTS of stuff that is more unique and opinionated, as an example.

PR #88, opened 2023-07-29
fix: set rpm-ostree updates as persistent

### LINKED ISSUE #87: Automated rpm-ostree updates never run
The changes in #84 have broken automated updated of ~~flatpaks and~~ rpm-ostree. 

The PR says "If the host machine is suspended or powered down during that time, the timers will execute on next boot." That is incorrect, as the timers are NOT set as Persistent, so unless you machine is up at 4AM (mine isn't), they are never run.

EDIT: the issue is only on the rpm-ostree timer. Flatpak update timer is set as persistent, as expected.

p5:
Resolves #87

According to #87, automated updates have not been running since the systemd service timer is not persistent.  Therefore, the machines must be running at 4am for the systemd service to be triggered.
This PR switches the service to persistent so it runs on next boot.

> Persistent=
Takes a boolean argument. If true, the time when the service unit was last triggered is stored on disk. When the timer is activated, the service unit is triggered immediately if it would have been triggered at least once during the time when the timer was inactive. Such triggering is nonetheless subject to the delay imposed by RandomizedDelaySec=. This is useful to catch up on missed runs of the service when the system was powered down. Note that this setting only has an effect on timers configured with OnCalendar=. Defaults to false.

> Use systemctl clean --what=state … on the timer unit to remove the timestamp file maintained by this option from disk. In particular, use this command before uninstalling a timer unit. See [systemctl(1)](https://www.freedesktop.org/software/systemd/man/systemctl.html#) for details.

### DIFF
--- a/files/etc/systemd/system/rpm-ostreed-automatic.timer
+++ b/files/etc/systemd/system/rpm-ostreed-automatic.timer
@@ -6,5 +6,7 @@ ConditionPathExists=/run/ostree-booted
 [Timer]
 OnCalendar=*-*-* 4:00:00
 
+Persistent=true
+
 [Install]
 WantedBy=timers.target
--- a/files/etc/systemd/system/rpm-ostreed-automatic.timer
+++ b/files/etc/systemd/system/rpm-ostreed-automatic.timer
@@ -5,7 +5,6 @@ ConditionPathExists=/run/ostree-booted
 
 [Timer]
 OnCalendar=*-*-* 4:00:00
-
 Persistent=true
 
 [Install]


bsherman (changes requested):
@p5 thank you for catching this! Yes, clearly a mistake made as I intended for this timer unit to be the same as the flatpak update timers.

Super minor request: to match the flatpak timers formatting, would you please remove the empty line between "OnCalendar" and "Persistent"?

Thank you!

bsherman (approved):
Thank you!

p5:
@bsherman 
Done - [6045be3](https://github.com/ublue-os/config/pull/88/commits/6045be306c2226e1b3e4bdcb4eef8b1fa98e1793)

Sorry, I was almost certain I saw the same blank line elsewhere in the repo/org, so that's why I included it in this PR.  I can't find it now though so I must have been imagining it.

p5:
Ah, it was GitHub's UI making it look like a blank line!
image

Issue #90, opened 2023-08-09
Automatic rpm-ostree update service fails to run


fiftydinar:
I always consistently got this error, no matter if it's 04:00AM or other time since around 10 days ago:

error: Creating importer: Failed to invoke skopeo proxy method OpenImage: remote error: pinging container registry ghcr.io: Get "https://ghcr.io/v2/": dial tcp: lookup ghcr.io: Temporary failure in name resolution

All people which tried to help me told me to check if DNS functions properly on my router.
I checked it & here are the results: router + Cloudflare DNS work properly, everything, including ghcr.io pings positively.

So, that is not the issue.

I & some other user were the only ones with the issue, while other user (Bigpod) verified to me that rpm automatic updates works in his case. Which assumes that most users did not have this issue unlike me & silverblueperson (Discord username). Than, after around 5-6 days, other user complained with the same issue as me.

I am not sure how to log this, as this is the only error I get from the systemd service.

I tried to troubleshoot it by manually running the rpm-ostree upgrade command & by manually running the systemd automatic rpm-ostree upgrade service. Both results went positive...

It is very weird.

castrojo:
I don't have time to dig into this today but we set the value to stage in this service and make no other changes, so not sure what is going on here. This service has been reliable for me for years so maybe there was a new version or something changed?

bsherman:
> I don't have time to dig into this today but we set the value to stage in this service and make no other changes, so not sure what is going on here. This service has been reliable for me for years so maybe there was a new version or something changed?

The change was in https://github.com/ublue-os/config/pull/84/files#diff-f1bc9c48a9bad1cf5a8198ee3434651663e8c497ffd9afdfcc915b0b2df382b7

The new timer works fine for anyone who leaves the computer on all night, or maybe after a resume from sleep, but sometimes, on a boot/resume from a machine which was off when the timer should have run, DNS resolution is not available. I think we may just need to add the following to the service.

```
Wants=network-online.target
After=network-online.target
```

At least, that's a good first step.

bsherman:
With the PR merged for this we can probably close this issue, but it's OK to wait a few for any feedback that it is/isn't working.

fiftydinar:
I can confirm that the issue is fixed!
![Screenshot from 2023-08-14 13-17-35](https://github.com/ublue-os/config/assets/65243233/19a36f5b-788a-479c-b8ee-d23650889958)

lorduskordus:
[...]

PR #93, opened 2023-08-12
fix: use rpm-ostreed-automatic.service drop-in to wait for network

fiftydinar:
For fixing this issue:
https://github.com/ublue-os/config/issues/90

"Temporary failure in name resolution"

Override the service instead of making the changes directly.


bsherman (changes requested):
I don't understand why this is adding a new file instead of modifying the existing service file. 

I may truly be mistaken on this if we only currently customize the timer and not the service. 

I am on mobile for a while and hard to review but I'll will check this when I'm back at the computer in a bit.

akdev1l:
mm two thoughts:

1. does this change actually belong upstream? https://github.com/coreos/rpm-ostree/blob/main/src/daemon/rpm-ostreed-automatic.service - if the service needs that then we should also think about making a PR to rpm-ostree itself
2. We should use a systemd drop-in to avoid clobbering the whole file. We can create a file in `/usr/etc/systemd/system/rpm-ostreed-automatic.service.conf.d/override.conf` with just this content:

```
[Unit]
Wants=network-online.target
After=network-online.target
```

this would have the effect of only setting these two properties while leaving everything else the same as upstream

(also @bsherman I checked and indeed we only have a `.timer` file currently)

bsherman:
> (also @bsherman I checked and indeed we only have a `.timer` file currently)

Thanks for confirming, back and desk and I agree. 


>    2. We should use a systemd drop-in to avoid clobbering the whole file.

Hmm... now I wonder why I didn't use a system drop-in for the timer itself... that seems like I messed up.

akdev1l (approved):

bsherman (approved):

PR #95, opened 2023-08-12
fix: use systemd drop-in to override rpm-ostreed-automatic timer

bsherman:
A better way to address this part of #26  without clobbering the timer.


files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.conf.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Timer]
```
akdev1l, line 1: this file needs to be in `rpm-ostreed-automatic.timer.d` not in `rpm-ostreed-automatic.timer.conf.d`

akdev1l (changes requested):

files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.conf.d/override.conf
bsherman: thank you!

akdev1l (approved):

PR #96, opened 2023-08-13
fix: run flatpak two minutes after boot

### LINKED ISSUE #89: Missed flatpak auto-update timers trigger before dbus session exists
After moving from GNOME to Hyprland, I've noticed that my flatpaks are no longer automatically updating. Previously, GNOME Software seemed to handle Flatpak updates. After removing GNOME Software, I expected the [update timers mentioned in the docs](https://universal-blue.org/faq/?h=update#how-do-i-configure-automatic-updates) to continue keeping flatpaks up to date, though this doesn't seem to be the case - manually running `flatpak update` every once in a while results in a significant list of updates from the previous few days.

While the system update timer is triggering, the service it starts is consistently failing. Here's the relevant part of the log from `systemctl status flatpak-system-update.service`:
```
Aug 08 17:03:11 fedora flatpak[1284]: error: Cannot autolaunch D-Bus without X11 $DISPLAY
Aug 08 17:03:11 fedora systemd[1]: flatpak-system-update.service: Main process exited, code=exited, status=1/FAILURE
Aug 08 17:03:11 fedora systemd[1]: flatpak-system-update.service: Failed with result 'exit-code'.
lines 1-17/17 (END)
```

Something to note here is that 17:03:11 was around the time SDDM started right after booting up my system, and I didn't log in for a few minutes after that. It seems like since the system was off for its scheduled time, the update service is being ran immediately after boot instead. It then fails because the update command requires dbus, which isn't present until after I log in.

It also seems like flatpak updates with systemd failing due to dbus has been raised as an issue on flatpak's repo before: https://github.com/flatpak/flatpak/issues/4345, to which the response was:

> This is nothing to do with Flatpak.
> If you want to run programs in a systemd system service that behave as though they are a tiny login session for D-Bus purposes, you should probably either use dbus-run-session(1), or if your OS includes /usr/lib/systemd/user/dbus.socket, register the systemd service as a login session by setting its PAMName= to a PAM stack (see /etc/pam.d or similar) that includes libpam_systemd.
https://github.com/flatpak/flatpak/issues/4345#issuecomment-883522502

However, `dbus-run-session -- flatpak update` seems to fail for me, as flatpak errors with message `A connection to the bus can't be made`

I think setting PAMName for the service is something worth investigating, along with the other solution I saw in a linked issue:  have the update service be a user service instead of system one, which would ensure that it doesn't start until the user actually logs in and dbus is running. `flatpak-user-update.service` is already a user service, and it does run successfully once the user signs in, so it looks like moving `flatpak-system-update.service` into `/usr/lib/systemd/user` would fix the issue.

I'm not sure if a time delay would sufficiently solve the issue, as it's not guaranteed that the user will be signed in within, say, 60 seconds of the system booting.

bsherman:
This should fix #89 by ensuring that flatpak update timers always run at least two minutes after system boot.

Also, randomized delay was added for both flatpak update and rpm-ostreed-automatic timers to prevent the download and system load from always hitting at exactly the same time.

### DIFF
--- a/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf
+++ b/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf
@@ -1,3 +1,4 @@
 [Timer]
+RandomizedDelaySec=10m
 OnCalendar=*-*-* 4:00:00
 Persistent=true
--- a/files/usr/lib/systemd/system/flatpak-system-update.timer
+++ b/files/usr/lib/systemd/system/flatpak-system-update.timer
@@ -3,6 +3,8 @@ Description=Flatpak Automatic Update Trigger
 Documentation=man:flatpak(1)
 
 [Timer]
+RandomizedDelaySec=10m
+OnBootSec=2m
 OnCalendar=*-*-* 4:00:00
 Persistent=true
 
--- a/files/usr/lib/systemd/user/flatpak-user-update.timer
+++ b/files/usr/lib/systemd/user/flatpak-user-update.timer
@@ -3,6 +3,8 @@ Description=Flatpak Automatic Update Trigger
 Documentation=man:flatpak(1)
 
 [Timer]
+RandomizedDelaySec=10m
+OnBootSec=2m
 OnCalendar=*-*-* 4:00:00
 Persistent=true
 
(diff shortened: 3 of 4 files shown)


bsherman:
I munged the changelog a bit as it seemed silly to bump the version 3 times today :smile:

akdev1l (approved):

PR #97, opened 2023-08-14
Update main.just

jxan13:
https://docs.fedoraproject.org/en-US/fedora-silverblue/troubleshooting/#_unable_to_add_user_to_group

might be easier for a new user to Silverblue if they layer libvirt packages.

### DIFF
--- a/build/ublue-os-just/main.just
+++ b/build/ublue-os-just/main.just
@@ -66,3 +66,8 @@ update:
 fixscreenshare:
   cp /usr/share/applications/org.kde.xwaylandvideobridge.desktop $HOME/.config/autostart/
   
+# Add user to libvirt group after layering libvirt packages
+usermod-libvirt:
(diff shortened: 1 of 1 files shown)


build/ublue-os-just/main.just
```diff
@@ -66,3 +66,8 @@ update:
 fixscreenshare:
   cp /usr/share/applications/org.kde.xwaylandvideobridge.desktop $HOME/.config/autostart/
   
+# Add user to libvirt group after layering libvirt packages
+usermod-libvirt:
```
p5, line 70: Can't we ship polkit rules that do the same thing?  

Is it better adding the user to the libvirt group or adding the necessary polkit rules?

jxan13: That would be way better than my one off approach. I am hoping that someone can fill in the gaps between mutable and immutable. In the mean time, using just gives a temporary way to get that done.

PR #98, opened 2023-08-17
Just target for sudo NOPASSWD

rothgar:
Added the target and an alias

This is how it shows up on my system
```
    sudo-nopass              # Allow current user to sudo without a password
    iamroot                  # alias for `sudo-nopass`
```

### DIFF
--- a/build/ublue-os-just/main.just
+++ b/build/ublue-os-just/main.just
@@ -66,3 +66,7 @@ update:
 fixscreenshare:
   cp /usr/share/applications/org.kde.xwaylandvideobridge.desktop $HOME/.config/autostart/
   
+# Allow current user to sudo without a password
+alias iamroot := sudo-nopass
+sudo-nopass:
+  echo "$USER ALL=(ALL) NOPASSWD:ALL" | sudo tee /etc/sudoers.d/${USER}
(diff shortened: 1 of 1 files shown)


akdev1l (commented):
This isn’t really a safe thing to do. We shouldn’t encourage anyone to have passwordless sudo imo. It weakens security far too much

EyeCantCU:
I agree with @akdev1l on this one. I think things that directly weaken security, convenient or not, should be excluded. But others may have interesting insight here that I don't

castrojo:
Semi-related, if we documented how to do easier overrides locally then that might mitigate this: https://github.com/ublue-os/config/issues/100

For example  if I wanted to add `brew update` to my `just update` command I could set that up in `~/.justfile` so that I can have the things I want but still get the system updates. This would let people mix and match!

bigpod98 (commented):
Im not a fan of this, i dont think we should be encuraging such insecure practices

rothgar:
np, just thought I'd offer it for convenience. I'm assuming you wouldn't want a target that adds a user to the docker group either.

castrojo:
I'll happily accept the user->docker group one in bluefin!

https://github.com/ublue-os/bluefin/blob/main/usr/share/ublue-os/just/custom.just

bigpod98:
Yea that one is acceptable in my opinion as well

akdev1l:
mm keep in mind membership into a rootful docker daemon is equivalent to root, it’s trivial to get root with that access: `docker run —privileged —rm -it -c /:/run/host fedora:latest`

rothgar:
that's exactly what I was suggesting. Being in the docker group won't require a password to gain root access. The same security risk as sudo with NOPASSWD

PR #107, opened 2023-08-31
fix(just): add i915 to modprobe.blacklist

EternalWinter:
If you have an nvidia GPU and a CPU with igpu inside you won't be able to start your system probably due to driver/module conflicts?
One possible workaround is to set ```i915.modeset=0``` or ```module.blacklist=i915``` to your grub config as described below.

Some of the research on this is described [here](https://askubuntu.com/questions/1483916/ubuntu-with-disk-encryption-doesnt-show-me-encryption-password-dialog-at-startu/1483936#1483936) .

### DIFF
--- a/build/ublue-os-just/nvidia.just
+++ b/build/ublue-os-just/nvidia.just
@@ -3,6 +3,7 @@ set-kargs-nvidia:
     rpm-ostree kargs \
         --append-if-missing=rd.driver.blacklist=nouveau \
         --append-if-missing=modprobe.blacklist=nouveau \
+        --append-if-missing=rd.driver.blacklist=i915 \
         --append-if-missing=modprobe.blacklist=i915 \
         --append-if-missing=nvidia-drm.modeset=1 \
         --delete=nomodeset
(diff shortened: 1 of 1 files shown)


bsherman:
Sorry, but this PR can't go forward. It assumes that all user who may require the nvidia driver would want to disable their intel iGPU, if it exists.

This is a non-starter for any hybrid laptop which primarily uses iGPU with an NVIDIA dGPU for 3D rendering. It would completely disable the display output.

I'm not sure if you are the original poster on the askubuntu thread, but in their case where they have a "RTX 3070TI and 12th Gen Intel® Core™ i9-12900K × 24"... I'd suggest disabling the iGPU in the system UEFI/BIOS settings. 

This is how I've addressed odd problems with my iGPU and discrete AMD/NVIDIA GPUs on my desktop. Its unfortunately not a terribly uncommon problem.

EternalWinter:
> Sorry, but this PR can't go forward. It assumes that all user who may require the nvidia driver would want to disable their intel iGPU, if it exists.
> 
> This is a non-starter for any hybrid laptop which primarily uses iGPU with an NVIDIA dGPU for 3D rendering. It would completely disable the display output.
> 
> I'm not sure if you are the original poster on the askubuntu thread, but in their case where they have a "RTX 3070TI and 12th Gen Intel® Core™ i9-12900K × 24"... I'd suggest disabling the iGPU in the system UEFI/BIOS settings.
> 
> This is how I've addressed odd problems with my iGPU and discrete AMD/NVIDIA GPUs on my desktop. Its unfortunately not a terribly uncommon problem.

Oh, thank you so much! Yes, i am original poster on the askubuntu.

EternalWinter:
@bsherman could i ask you few more questions?

PR #108, opened 2023-09-01
feat(just): add Fedora shortcut

castrojo:


bobslept (commented):
It just belongs there. good catch

KyleGospo (approved):

PR #115, opened 2023-09-14
feat(just): move mullvad vpn shortcut to vpn.just

Renner0E:
Made a little more improvements to the script
Should I remove signing keys from the keyring or should they stay? Is anyone bothered by this?


castrojo:
Sorry it's taken a while to review, just letting you know we want this but it's a matter of time to review, thanks!

PR #118, opened 2023-09-20
fix: Split justfiles into multiple numbered files to fix issue with default command on other releases.

KyleGospo:
Cleaner thanks to commands now being categorized, and additionally fixes an issue where just files are loaded alphabetically and the default command can change as new files are added.

Brings in comments and listing improvements made at Bazzite.


castrojo:
cc @ublue-os/approver for feedback.

I think we need to account for recreating the existing user's `~/.justfile` so that they have a migration path. Probably a one shot service unit?

KyleGospo:
Alternative option, do we have the ability to define what's in `~/.justfile` in a system level file and avoid having to use files in home at all?

p5:
> Alternative option, do we have the ability to define what's in `~/.justfile` in a system level file and avoid having to use files in home at all?

For my image, I have aliased `eternal` (for Eternal OS :smile:) to be `just --unstable --justfile /usr/share/eternal/justfile`.
So by default, users (just me) needs to run `eternal` and uses the justfiles installed in the system.
The `/usr/share/eternal/justfile` file contains only includes which are added on each image during build - 
```
RUN echo "!include /usr/share/eternal/lumina.just" >> /usr/share/eternal/justfile && \
  echo "!include /usr/share/eternal/lumina-setup.just" >> /usr/share/eternal/justfile
```

build/ublue-os-just/40-nvidia.just
bsherman, line 1: This reorg seems to always include the `nvidia.just` (now `40-nvidia.just`) file... even on non nvidia images?

If so, I'd suggest renaming the 3 nvidia-specific recipes to:
- nvidia-set-kargs
- nvidia-test-cuda
- nvidia-setup-firefox-vaapi

This will group the nvidia recipes in the output of `just -l` (eg, default recipe) for easier ignoring by amd/intel users.

bsherman (changes requested):
I'm requesting these changes, but let me know if I'm off base.

build/ublue-os-just/40-nvidia.just
KyleGospo: Done with some other minor changes

joefiorini:
I like @p5's idea of using an alias, except that might require a little more effort to be supported across different shells. is there appetite for having a fairly simple cli that wraps just and sets up paths on its own?

In the meantime, what about having a master justfile in `/usr/share/ublue-os` that includes all of the numbered files in whatever order you want? Then in ~/.justfile it becomes:

```
...
!include /usr/share/ublue-os/master.justfile
...
```

You still have to deal with the user being able to edit ~/.justfile, but its a start 😉

KyleGospo:
> I like @p5's idea of using an alias, except that might require a little more effort to be supported across different shells. is there appetite for having a fairly simple cli that wraps just and sets up paths on its own?
> 
> In the meantime, what about having a master justfile in `/usr/share/ublue-os` that includes all of the numbered files in whatever order you want? Then in ~/.justfile it becomes:
> 
> ```
> ...
> !include /usr/share/ublue-os/master.justfile
> ...
> ```
> 
> You still have to deal with the user being able to edit ~/.justfile, but its a start 😉

Latest commit does the master justfile as /usr/share/ublue-os/justfile, now we just need the .justfile editor as mentioned

build/ublue-os-just/40-nvidia.just
KyleGospo: To add to this, we're planning on shipping a standardized image info file for ublue-update to use, so these can be modified to read it and warn the user that they're not on a Nvidia image instead of doing anything.

KyleGospo:
This now depends on this landing first:
https://github.com/ublue-os/main/pull/350

build/ublue-os-just/40-nvidia.just
KyleGospo: @bsherman Rethinking this, since not every image is going to ship that file, how about we instead look for the presence of nvidia libraries and throw an error if they aren't there? That makes it generic enough to work in any image no matter what's done to it.

bsherman: Don't all images ship it though?

Today they seem to do so... I run a derivative of `silverblue-main:38` and it contains the file: `/usr/share/ublue-os/just/nvidia.just`

bsherman: I'm fine if the nvidia-specific just recipies are more intelligent though.

EyeCantCU (approved):
Tested, can confirm things are working as intended!

build/ublue-os-just/40-nvidia.just
EyeCantCU: We can ship the script for obtaining image info in main and work around that. Sounds like it would be especially useful here

KyleGospo: > Don't all images ship it though?
> 
> Today they seem to do so... I run a derivative of `silverblue-main:38` and it contains the file: `/usr/share/ublue-os/just/nvidia.just`

That part's fine, I mean a check in nvidia.just that looks for /usr/lib/\*nvidia\* and displays a helpful message about you probably not being on a nvidia image instead of attempting to set kargs.

castrojo (approved):

bsherman (approved):
Looks great! Good improvements here!

PR #122, opened 2023-09-24
Fix #117 (flatpak update failing)

### LINKED ISSUE #117: flatpak-system-update.service: start-post operation timed out
For the past week or so, I have noticed that flatpaks did not auto update. I decided to look at the systemd logs and found the following:

```
× flatpak-system-update.service - Flatpak Automatic Update
     Loaded: loaded (/usr/lib/systemd/system/flatpak-system-update.service; static)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: timeout) since Tue 2023-09-19 14:15:22 EEST; 1h 53min ago
TriggeredBy: ● flatpak-system-update.timer
       Docs: man:flatpak(1)
    Process: 2390 ExecStartPre=/usr/bin/flatpak --system uninstall --unused -y --noninteractive (code=exited, status=0/SUCCESS)
    Process: 2396 ExecStart=/usr/bin/flatpak --system update -y --noninteractive (code=exited, status=0/SUCCESS)
    Process: 2397 ExecStartPost=/usr/bin/flatpak --system repair (code=exited, status=1/FAILURE)
   Main PID: 2396 (code=exited, status=0/SUCCESS)
        CPU: 28.256s

Sep 19 14:15:21 fedora flatpak[2397]: [49/145] Verifying flathub:runtime/org.freedesktop.Sdk.Locale/x86_64/22.08…
Sep 19 14:15:21 fedora flatpak[2397]: [50/145] Verifying flathub:runtime/org.kde.PlatformTheme.QGnomePlatform/x86_64/5.15-22.08…
Sep 19 14:15:21 fedora flatpak[2397]: [51/145] Verifying flathub:runtime/org.freedesktop.Platform.VAAPI.Intel/x86_64/21.08…
Sep 19 14:15:22 fedora flatpak[2397]: [52/145] Verifying flathub:runtime/org.freedesktop.Platform.GL.default/x86_64/21.08…
Sep 19 14:15:22 fedora systemd[1]: flatpak-system-update.service: start-post operation timed out. Terminating.
Sep 19 14:15:22 fedora flatpak[2397]: [6B blob data]
Sep 19 14:15:22 fedora systemd[1]: flatpak-system-update.service: Control process exited, code=exited, status=1/FAILURE
Sep 19 14:15:22 fedora systemd[1]: flatpak-system-update.service: Failed with result 'timeout'.
Sep 19 14:15:22 fedora systemd[1]: Failed to start flatpak-system-update.service - Flatpak Automatic Update.
Sep 19 14:15:22 fedora systemd[1]: flatpak-system-update.service: Consumed 28.256s CPU time.
```

I then ran the commands from the unit manually, and each of them completed just fine, updating all the flatpaks. From my understanding, this is caused by the fact that ```flatpak --system repair``` is taking way too long, and getting terminated by systemd, which thinks it froze and is doing nothing.

At first I wanted to suggest simply increasing the timeout for ```ExecStartPost```, but as I was looking for the option in the [systemd service specification](https://www.freedesktop.org/software/systemd/man/systemd.service.html), I found the following:

> Note that ExecStartPre= may not be used to start long-running processes. All processes forked off by processes invoked via ExecStartPre= will be killed before the next service process is run.

So it seems that they don't recommend using this option to run commands that may take a lot of time, and therefore, there is no way to customize the timeout for it - there are only ```TimeoutStartSec```, ```TimeoutStopSec```, and ```TimeoutAbortSec``` (could this be dealt with by changing the service type? I don't know and haven't found a type that alters the behavior of ```ExecStartPost```, either way that would be closer to a workaround rather than an actual fix).

The only solution I see right now is starting all 3 commands in ```ExecStart```, possibly by creating a small shell script (the downside is that this makes debugging more difficult if all 3 commands are in one place? They were probably put in 3 different places for a reason after all).

If you approve of my "just put them all in one place" approach, then I could probably make a PR, if not, let's discuss how we can fix this :)

ArtikusHG:
This PR fixes issue #117, by making all the commands in the ```flatpak-system-update``` service run from ```ExecStart=```, which prevents them from being automatically killed from time to time.

### DIFF
--- a/files/usr/lib/systemd/system/flatpak-system-update.service
+++ b/files/usr/lib/systemd/system/flatpak-system-update.service
@@ -5,7 +5,5 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=simple
-ExecStartPre=/usr/bin/flatpak --system uninstall --unused -y --noninteractive
-ExecStart=/usr/bin/flatpak --system update -y --noninteractive
-ExecStartPost=/usr/bin/flatpak --system repair
+Type=oneshot
+ExecStart=/usr/bin/flatpak --system uninstall --unused -y --noninteractive ; /usr/bin/flatpak --system update -y --noninteractive ; /usr/bin/flatpak --system repair
--- a/files/usr/lib/systemd/user/flatpak-user-update.service
+++ b/files/usr/lib/systemd/user/flatpak-user-update.service
@@ -5,7 +5,5 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=simple
-ExecStartPre=/usr/bin/flatpak --user uninstall --unused -y --noninteractive
-ExecStart=/usr/bin/flatpak --user update -y --noninteractive
-ExecStartPost=/usr/bin/flatpak --user repair
+Type=oneshot
+ExecStart=/usr/bin/flatpak --user uninstall --unused -y --noninteractive ; /usr/bin/flatpak --user update -y --noninteractive ; /usr/bin/flatpak --user repair


ArtikusHG:
It'd be really cool if we could merge this today so that it lands on tomorrow's image :)

bsherman (changes requested):
Would you please make the same changes to the flatpak-user-update.service file?

ArtikusHG:
Oh darn, I totally forgot about that one. One moment!

ArtikusHG:
@bsherman done

bsherman (approved):
Looks good! Let's try it!

PR #127, opened 2023-09-28
feat: Add OpenTabletDriver udev rules pulled from their latest GitHub release

KyleGospo:


bsherman:
I think this is just missing a changelog update for the ublue-os-udev-rules RPM and should be good to go!

KyleGospo:
> I think this is just missing a changelog update for the ublue-os-udev-rules RPM and should be good to go!

Thanks, that's done now

castrojo (approved):

PR #130, opened 2023-10-02
feat(just): add fwupd firmware updates to justfile

Renner0E:
Or is it maybe better to have it separate and call the shortcut like update-firmware?


bsherman:
> Or is it maybe better to have it separate and call the shortcut like update-firmware?

I'd personally prefer a distinct `update-firmware` recipe, as I tend to update packages alot and firmware more tentatively.

bsherman (approved):
LGTM

PR #132, opened 2023-10-05
feat(just): add kernel parameters for overclocking AMD GPUs

Renner0E:
This is a kernel parameter change [required to overclock your AMD GPU.](https://wiki.archlinux.org/title/AMDGPU#Overclocking)
This is necessary to make programs like [CoreCtrl](https://gitlab.com/corectrl/corectrl) have full functionality.

### DIFF
--- a/build/ublue-os-just/41-amd.just
+++ b/build/ublue-os-just/41-amd.just
@@ -1,0 +1,12 @@
+# vim: set ft=make :
+# Set needed kernel parameters for Overclocking AMD GPUs
+amd-set-oc-kargs:
+  #!/usr/bin/env bash
+  if lsmod | grep -wq "amdgpu"; then
+    rpm-ostree kargs \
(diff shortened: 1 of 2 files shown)


bsherman:
Interesting... I'm a little hesitant to include aids to overclocking here as it would be included in all our `*-main` images... and I don't know the risks or implications. Plus, no one is really hindered from doing this today as it's just running rpm-ostree kargs.

@KyleGospo any thoughts if you'd want to add this to `60-custom.just` in bazzite?

KyleGospo:
> Interesting... I'm a little hesitant to include aids to overclocking here as it would be included in all our `*-main` images... and I don't know the risks or implications. Plus, no one is really hindered from doing this today as it's just running rpm-ostree kargs.
> 
> @KyleGospo any thoughts if you'd want to add this to `60-custom.just` in bazzite?

Already there as part of the corectrl installer

PR #135, opened 2023-10-13
feat: add ujust, add chsh just task

b-:
This adds a little shim, `/usr/bin/ujust`, to run, `just --unstable --justfile /usr/share/ublue-os/justfile`.

I think this is a much cleaner way to run our `just` tasks than putting a `~/.justfile` in the home directory, for a number of reasons. One of them is that this does not require aliasing `just=just --unstable`, and another is that a user can actually have their own `~/.justfile` with no regard for including our global one.

I didn't remove the `just` aliases from the shell dotfiles, because I'm not sure what breaking changes that might cause -- documentation, for sure, but also does yafti or anything else depend on that?

~~Also to be honest I'm not sure where the default just alias comes from. I didn't actually look yet :P~~ _the alias comes from this repo_

I did this because I realized how much duplicate code and room for errors I had in the shell change tasks I rewrote -- I think it's much better to write a task with an argument, `just chsh /usr/bin/newshell`, and then call that from other tasks that need to change the shell. However, I saw the limitations of using shell aliases to add the `--unstable` flag quite quickly -- my sub-just-invocations became longer than I'd like them to be. See https://just.systems/man/en/chapter_40.html and note how much nicer that is than `just --unstable --file /usr/share/ublue-os/justfile`...

I'd definitely like to see others' thoughts on this! Feel free to ping me on Discord @bri9.

PR is draft because I'd like others to test it just in case I messed something up here...

@castrojo suggested this go in this repo. I've never built an RPM before and I'm not certain I did it right, so I'm sort of crossing my fingers and teaching myself how to test etc. as I go along...

--bri

### DIFF
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -1,5 +1,8 @@
 # vim: set ft=make :
 
+uid := `id -u`
+shell := `grep :$(id -u): /etc/passwd | cut -d: -f7`
+
 set allow-duplicate-recipes
 set ignore-comments
 
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -10,6 +10,21 @@ _default:
 bios:
   systemctl reboot --firmware-setup
 
+# Change the user's shell
+chsh new_shell:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  if [ "{{shell}}" = "{{new_shell}}" ] ; then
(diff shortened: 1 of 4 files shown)


b-:
I think I properly built the RPM, so I marked this ready for review, but I have to be clear it's really "ready for review" as I've never build an RPM before!

Zoialord (approved):
Looks good and def a good cleanup

KyleGospo:
Very good cleanup

KyleGospo (approved):

PR #138, opened 2023-10-20
feat(udev): add Sunshine udev rules (fix #119)

### LINKED ISSUE #119: Include Sunshine udev rules
### Describe the package

I'd like to request the package `[Sunshine](https://github.com/LizardByte/Sunshine/tree/master)` because it allows streaming to any device that has access to a Moonlight client with support for AMD, Intel, and Nvidia GPUs for hardware encoding.

It is the best way I've found to stream to any device I want. The stream quality is good and devices with Moonlight are plentiful (i.e. [Nintendo Switch Moonlight Homebrew](https://github.com/XITRIX/Moonlight-Switch)).

One issue I found and I'm not sure it affects Sunshine so much as it's a Moonlight client issue but it's worth mentioning. When the client device performs Wi-Fi scanning it causes Moonlight to hitch on Linux/GNOME. The only way I found to fix that was to set a BSSID on the client to prevent network scans. I'm unsure if this would affect the server-side in the same way.

It is available as a Flatpak however there is some setup that needs to be performed so it'd be ideal if it was baked in. Though as it is available in other packages it's just a nice to have.

### Information on the package

Installed Packages
Name         : sunshine
Version      : 0.20.0
Release      : 1
Architecture : x86_64
Size         : 48 M
Source       : sunshine-0.20.0-1.src.rpm
Repository   : @System
From repo    : @commandline
Summary      : Sunshine is a self-hosted game stream host for Moonlight.
URL          : https://app.lizardbyte.dev
License      : unknown
Description  : DESCRIPTION
             : ===========
             : 
             : This is an installer created using CPack (https://cmake.org). No
             : additional installation instructions provided.

### Image

Bazzite

ArtikusHG:
This PR fixes issue #119 by adding Sunshine udev rules. These don't seem to be hardware-specific or regularly updating (as per [documentation](https://github.com/LizardByte/Sunshine/blob/master/docs/source/about/usage.rst)), so it's fine to include them as a static file.

### DIFF
--- a/files/etc/udev/rules.d/85-sunshine.rules
+++ b/files/etc/udev/rules.d/85-sunshine.rules
@@ -1,0 +1,1 @@
+KERNEL=="uinput", SUBSYSTEM=="misc", OPTIONS+="static_node=uinput", TAG+="uaccess"
--- a/rpmspec/ublue-os-udev-rules.spec
+++ b/rpmspec/ublue-os-udev-rules.spec
@@ -1,7 +1,7 @@
 Name:           ublue-os-udev-rules
 Packager:       ublue-os
 Vendor:         ublue-os
-Version:        0.5
+Version:        0.6
 Release:        1%{?dist}
 Summary:        Additional udev files for device support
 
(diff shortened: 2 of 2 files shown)


castrojo (approved):
Thanks!

ArtikusHG:
Thanks for merging quickly! Also, we can now close #119 :)

PR #139, opened 2023-10-24
fix: REVERT feat(udev): add Sunshine udev rules (fix #119)

### LINKED ISSUE #119: Include Sunshine udev rules

bsherman:
This reverts commit 4d0e776a0537cff8e9c15165e88137bea78e8ed6.

/usr/lib/udev/rules.d/85-sunshine.rules conflicts with the sunshine RPM itself, preventing the install of that package.

I don't think the intent of this PR was to prevent users from using the sunshine RPM.

@ArtikusHG - FYI - I think we need to improve on this.

### DIFF
--- a/files/etc/udev/rules.d/85-sunshine.rules
+++ b/files/etc/udev/rules.d/85-sunshine.rules
@@ -1,1 +1,0 @@
-KERNEL=="uinput", SUBSYSTEM=="misc", OPTIONS+="static_node=uinput", TAG+="uaccess"
--- a/rpmspec/ublue-os-udev-rules.spec
+++ b/rpmspec/ublue-os-udev-rules.spec
@@ -1,7 +1,7 @@
 Name:           ublue-os-udev-rules
 Packager:       ublue-os
 Vendor:         ublue-os
-Version:        0.6
+Version:        0.5
 Release:        1%{?dist}
 Summary:        Additional udev files for device support
 
@@ -45,9 +45,6 @@ cp %{buildroot}%{_datadir}/%{VENDOR}/{%{sub_name}/etc/udev/rules.d,game-devices-
 
 
 %changelog
-* Fri Oct 20 2023 ArtikusHG <24320212+ArtikusHG@users.noreply.github.com> - 0.6
-- Add Sunshine udev rules
-
 * Thu Sep 28 2023 Kyle Gospodnetich <me@kylegospodneti.ch> - 0.5
 - Add OpenTabletDriver udev rules
 


bsherman:
Discord thread documenting the problem: https://discord.com/channels/1072614816579063828/1072617059265032342/1166181845264773240

> Hey folks, I've been using sunshine as a layered package so I can use the systemd scripts to automatically start it on login.
This PR seems to be causing me a problem: https://github.com/ublue-os/config/pull/138
When I update I now get the folloing
`error: Checkout sunshine-0.20.0-1.x86_64: Hardlinking 4e/843b6d65bd7666b282e019d1cd9ce8a8a5efc72f8be9ebc58080bcca6344f4.file to 85-sunshine.rules: File exists`
> 
> Is my only solution to switch to the flatpack in order to update my system?

RedTopper:
Thanks @bsherman for opening the report for me!

As an additional datapoint, it is not possible to remove the layered sunshine package, update, then re-layer. It produces the same error.

My current use-case for the RPM is to use the systemd rules to start and stop the service. I could likely re-work to use the flatpak instead, but it seems odd to make that the only option.

bsherman:
I'm working on a better solution than a simple revert.

bsherman:
Better solution: #140

PR #140, opened 2023-10-24
fix: rename Sunshine/OpenTabletDriver udev rules to avoid collissions

bsherman:
This renames the files 70-opentabletdriver.rules and and 85-sunshine.rules to add "-ublue" before ".rules" and increment the numbered prefix by one.

The "-ublue" ensures we'll never have a name collision which prevents their respective official RPMs from installing.

The increment attempts to assure that the official rules will take priority should the official RPM be installed.

FYI - @ArtikusHG / @KyleGospo

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,6 +1,6 @@
 FROM registry.fedoraproject.org/fedora:latest AS builder
 
-RUN dnf install --disablerepo='*' --enablerepo='fedora,updates' --setopt install_weak_deps=0 --nodocs --assumeyes rpm-build systemd-rpm-macros wget
+RUN dnf install --disablerepo='*' --enablerepo='fedora,updates' --setopt install_weak_deps=0 --nodocs --assumeyes rpm-build systemd-rpm-macros wget jq git
 
 ADD https://codeberg.org/fabiscafe/game-devices-udev/archive/main.tar.gz /tmp/ublue-os/rpmbuild/SOURCES/game-devices-udev.tar.gz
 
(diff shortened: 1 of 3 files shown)


Containerfile
```diff
@@ -11,14 +11,21 @@ ADD files/etc/udev/rules.d /tmp/ublue-os/udev-rules/etc/udev/rules.d
...
-| tr -d \" \
+| jq -r '.assets | sort_by(.created_at) | .[] | select (.name|test("opentabletdriver.*tar.gz$")) | .browser_download_url' \
 | wget -qi - -O /tmp/OpenTabletDriver/opentabletdriver.tar.gz && \
-tar -xvzf /tmp/OpenTabletDriver/opentabletdriver.tar.gz -C /tmp/OpenTabletDriver && \
-mv /tmp/OpenTabletDriver/opentabletdriver/etc/udev/rules.d/70-opentabletdriver.rules /tmp/ublue-os/udev-rules/etc/udev/rules.d/70-opentabletdriver.rules && \
+tar --strip-components=1 -xvzf /tmp/OpenTabletDriver/opentabletdriver.tar.gz -C /tmp/OpenTabletDriver && \
+mv /tmp/OpenTabletDriver/etc/udev/rules.d/70-opentabletdriver.rules /tmp/ublue-os/udev-rules/etc/udev/rules.d/71-opentabletdriver-ublue.rules && \
 rm -rf /tmp/OpenTabletDriver
 
+# Install Sunshin udev rules from their github repo
```
RedTopper, line 20:
```suggestion
# Install Sunshine udev rules from their github repo
```

RedTopper (approved):
Ah, thanks for figuring this one out so quickly!

PR #154, opened 2023-11-18
feat: add udev rule for an RCM'd Nintendo Switch to allow rootless payload injection

badlydrawnface:
Just to make having to reboot a jigged Switch easier, since I cannot get ns-usbloader to run at all.


KyleGospo (approved):

KyleGospo:
Looks good to me, Great hardware enablement.

PR #155, opened 2023-11-19
feat: check just syntax in build action (fix #152)

### LINKED ISSUE #152: Test just syntax in an action
We should make a check that runs `just` and then checks the syntax of the files in the PR so we don't land syntax errors: 

https://github.com/marketplace/actions/setup-just

ArtikusHG:
This PR makes the build action fail when ``just --fmt --check`` fails (#152).

**Note: currently, none of the justfiles we have here pass the formatting check. #156 must be merged first.**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,6 +38,21 @@ jobs:
           fi
           echo "alias_tags=${alias_tags[*]}" >> $GITHUB_OUTPUT
 
+      - name: Install just
+        id: install_just
+        uses: extractions/setup-just@v1
+
+      - name: Check just syntax
+        id: check_just_syntax
+        shell: bash
+        run: |
+          for file in build/ublue-os-just/*; do
+            if [ -f "$file" ]; then
+              echo $file
+              just --fmt --check --unstable -f $file || { exit 1; }
+            fi
+          done
+
       # Build image using Buildah action
       - name: Build Image
         id: build_image
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         id: check_just_syntax
         shell: bash
         run: |
-          for file in build/ublue-os-just/*; do
+          for file in build/ublue-os-just/*.just; do
             if [ -f "$file" ]; then
               echo $file
               just --fmt --check --unstable -f $file || { exit 1; }


castrojo:
Thanks for both of these, this will catch my dumb mistakes! 

I need someone more technical to review @ublue-os/approver @ublue-os/member

bobslept:
Oh sweet, no more broken just ending up on the images! Happened a few times now, big quality improvement!

If this works well and is being approved we for sure need this (slightly modified) action in framework and bluefin repo too.

.github/workflows/build.yml
```diff
@@ -38,6 +38,21 @@ jobs:
...
 
+      - name: Install just
+        id: install_just
+        uses: extractions/setup-just@v1
+
+      - name: Check just syntax
+        id: check_just_syntax
+        shell: bash
+        run: |
+          for file in build/ublue-os-just/*; do
```
KyleGospo, line 49: Should be *.just

KyleGospo (changes requested):

.github/workflows/build.yml
ArtikusHG: Oh yeah sorry, totally missed that! Will fix.

KyleGospo (approved):

KyleGospo:
Thank you for this!!

ArtikusHG:
@bobslept how many places do we need this in? If it's more than 2-3 it might make sense to not copy-paste the code but instead to make a separate repo with with this action, then we'll be able to just do something like:
```
- name: Check just syntax
  uses: ublue-os/check-just-syntax
```

fiftydinar:
@ArtikusHG startingpoint would definitely need this

ArtikusHG:
@fiftydinar it seems like we *do* need this in many places after all. I will make a separaye repo with an action that we can easily include tomorrow. If someone could make a repo for it under the ublue org and give me write access that'd help a lot!

castrojo:
https://github.com/ublue-os/just-action/ 

All set, check your emails! Thanks ya'll!

bobslept:
> @bobslept how many places do we need this in? If it's more than 2-3 it might make sense to not copy-paste the code but instead to make a separate repo with with this action, then we'll be able to just do something like:
> 
> ```
> - name: Check just syntax
>   uses: ublue-os/check-just-syntax
> ```

@ArtikusHG sure is. Would be neat if we can specify the directory where the just files are. Because those differ from repo to repo.

ArtikusHG:
@bobslept yes I'm planning to do that, the final thing will look something like:
```
- name: Check just syntax
  uses: ublue-os/just-action
  with:
    directory: build/ublue-os-just
```

bobslept:
> @bobslept yes I'm planning to do that, the final thing will look something like:
> 
> ```
> - name: Check just syntax
>   uses: ublue-os/just-action
>   with:
>     directory: build/ublue-os-just
> ```

That is perfect. Thank you so much! :partying_face:

ArtikusHG:
Please take a look at https://github.com/ublue-os/just-action/pull/1

PR #159, opened 2023-11-22
feat(just): Added homebrew docker container creation to 30-distrobox.…

dnkmmr69420:
…just

This would be useful if you want an instance of homebrew to be more isolated from your system


castrojo (approved):
Nice!

KyleGospo (approved):

PR #162, opened 2023-11-27
feat(just): More nix things added to 70-nix.just

dnkmmr69420:


build/ublue-os-just/70-nix.just
dnkmmr69420, line 1: This fixes the daemon not starting during boot

dnkmmr69420:
I made the nix systemd unit files real files since they don't start since they are symlinks

KyleGospo (approved):

p5 (approved):
LGTM

Issue #148, opened 2023-11-07
just has inconsistent naming for enable/disable scripts


bronson:
I see at least three styles of enable/disable scripts in `just --list`:

* disable-updates ⬌ enable-updates
* disable-broadcom-wl ⬌ use-broadcom-wl
* use-stock-gamepads ⬌ use-ublue-gamepads

Worth standardizing? Feel free to close if not!

castrojo:
This is related to: https://github.com/orgs/ublue-os/discussions/202

Basically we need someone to propose/reorg the justfiles to make sense since they kinda grew organically. And perhaps we need to add a little file with conventions to use or something like that.

HikariKnight:
I should in theory have some more freetime soon now in December so i will start having a look at reorging just, maybe shrink it down to toggles where we can.

HikariKnight:
after a brief discussion on discord, i will have a look at mocking together a bash script that will check if gum is installed and then either use gum or a basic bash dialog/prompt for toggles, that should at least get us started on cleaning up some of the just commands.

Currently the fallback might look something like this
![image](https://github.com/ublue-os/config/assets/2557889/79949242-803a-4620-8a46-c7fa184ce606)

HikariKnight:
i have a working concept now which works, bit unsure if a helper script for just and configs actually fit in ublue-os/config

invalid choices gets the output piped to STDERR and user gets asked again to pick an option, valid choices gets echoed to STDOUT and then exits, this should at least help us merge many commands into just toggle commands.
![image](https://github.com/ublue-os/config/assets/2557889/e92f4eea-36f7-4d78-9946-d0b80962a24e)

### Proposal for standardized naming scheme
- Toggles to be named `toggle-thing`
- Recipes installing/uninstalling/configuring things to be named `setup-thing` or `configure-thing` which then provides options for what the user wants to do, to avoid complexity we can also just split `setup-` and `configure-` to do their own thing.
- Recipes that only do 1 thing to be named `thing` like `ujust update`

castrojo:
Naming scheme sounds good to me!

HikariKnight:
I will wait for more responses until after December 29th before i start working on this in ublue-os/config.
I will open a PR draft on the 29th so changes can be discussed there while i work on it.

once config is done i will start branching out to the other repos (bazzite, bluefin, kinoite, etc) and take a look (unless someone pings me and tells me they will take care of it themselves)

Think that will be my plan forward regarding this at least.

PR #173, opened 2023-12-20
feat: add gum helper for choose dialogs

HikariKnight:
This should help start the process of addressing #148 as it will let us provide basic replacement for gum on systems that lack it with minimal options which in turn lets us shrink down the amount of just commands we have greatly.

Currently it only does choose dialogs, but before merging i want to implement the confirm dialog too.
Return values will be identical to what you would expect from gum so in the most basic use cases it will be a 1:1 drop in, assuming you do not need anything more fancy than the defaults from gum.

### DIFF
--- a/build/ublue-os-just/build.sh
+++ b/build/ublue-os-just/build.sh
@@ -9,6 +9,7 @@ mkdir -p /tmp/ublue-os/rpmbuild/SOURCES
 cp ${SCRIPT_DIR}/*.just /tmp/ublue-os/rpmbuild/SOURCES
 cp ${SCRIPT_DIR}/ublue-os-just.sh /tmp/ublue-os/rpmbuild/SOURCES
 cp ${SCRIPT_DIR}/ujust /tmp/ublue-os/rpmbuild/SOURCES
+cp ${SCRIPT_DIR}/ugum /tmp/ublue-os/rpmbuild/SOURCES
 
 rpmbuild -ba \
     --define '_topdir /tmp/ublue-os/rpmbuild' \
--- a/build/ublue-os-just/ublue-os-just.spec
+++ b/build/ublue-os-just/ublue-os-just.spec
@@ -1,7 +1,7 @@
 Name:           ublue-os-just
 Packager:       ublue-os
 Vendor:         ublue-os
-Version:        0.9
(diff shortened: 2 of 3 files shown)


HikariKnight:
Current implementation of `ugum` supports the `choose` and `confirm` verbs
if a command was to be `gum choose --cursor="| " "ace of spades" "king of hearts" "5 of diamond"` in a just recipe, this can now be replaced with `ugum choose --cursor="| " "ace of spades" "king of hearts" "5 of diamond"` and ugum will ignore any arguments starting with `-` and only focus on the the basics, which is providing a choice selection for the user, if `ugum` detects that the `gum` binary is present, it will pass all arguments to `gum` for a more fancy output.
The return values in STDOUT will match what you would expect from using `gum` if it was installed and it will return the option the user selected.

`ugum confirm` does the same for confirm dialogs and would be the equivalent to `gum confirm`.
To change the default prompt just supply one after `confirm` like `ugum confirm "Do you want to build a rocket ship?"`
exit code will be 0 if yes, 1 if no. This matches the output from `gum`

HikariKnight:
Is there any more changes i need to do to the spec file since it says it failed to build due to the ugum file missing in the log? :thinking:

bsherman (dismissed):
Seems reasonable.

HikariKnight:
fzf is now being used as fallback handler if gum is not available and will look like this
image

if no supported handler is provided in $MENU environment variable and fzf is not available, it will use basic bash to generate the choice selection and the confirm prompts

KyleGospo:
This looks great!!

HikariKnight:
Ok, as long as no more needed changes get requested, this is now feature complete for initial release.
There were some requests to use fzf after the PR got opened for review.
Since fzf is included in the images, making it a perfect alternative for gum for those images.

The generic bash alternatives are still there in case a user makes a custom image and removes fzf and gum for whatever reason.

Any more changes will be done at a later date with their own PR and if support for more menu handlers is desired then it can be added at a later date by people familiar with the said menu handlers (dmenu, rofi, wofi, rofi-wayland, etc)

KyleGospo (approved):

bsherman (approved):

PR #174, opened 2023-12-21
feat(udev): add udev rule for promptless charing threshold changes 

ispanos:
Allows user to modify the battery charge thresholds without password prompt. The rule was copied from [marcosdalvarez/thinkpad-battery-threshold-extension](https://gitlab.com/marcosdalvarez/thinkpad-battery-threshold-extension). Works only on ThinkPads.

### DIFF
--- a/files/etc/udev/rules.d/99-thinkpad-thresholds-udev.rules
+++ b/files/etc/udev/rules.d/99-thinkpad-thresholds-udev.rules
@@ -1,0 +1,7 @@
+# Change the permissions of the battery threshold attributes so that they can be modified by ordinary users
+# See https://gitlab.com/marcosdalvarez/thinkpad-battery-threshold-extension
+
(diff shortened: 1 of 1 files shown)


KyleGospo:
Looks good to me, are there any security implications of regular users being able to do this? Should we limit this to the wheel group?

ispanos:
@KyleGospo I suppose it could be used maliciously. I haven't tried it, but someone could potentially use it to render a laptop temporarily unusable without a charger, i.e. setting a very low charging threshold.
Setting a very low charging threshold (1-5), if it is not noticed soon could cause the battery to degrade rapidly.

KyleGospo:
I think that's fair, if you're up for it let's just add a restriction so this only works for wheel

ispanos (commented):
Allow only wheel to change Thinkpad charging thresholds. Did I do it properly?

KyleGospo (approved):

HikariKnight (approved):

PR #176, opened 2023-12-22
feat(just): added log messages shortcut from journalctl to 00-default.just

Renner0E:
I think it would be pretty handy if there was like a collection of shortcuts for log messages/crash reports.


castrojo (approved):

castrojo:
This ready for review?

KyleGospo (approved):

PR #181, opened 2024-01-01
feat(just): combine several just commands into toggles or choices

HikariKnight:
This patch, when complete will be the second step in addressing #148 now that we have `ugum` on all images which solved the need of a standardized way to select between options.

This is something that i will be chipping away at when i am able to.

Naming scheme for just recipes (draft) will be:
- `setup-` Used when the recipe provides a way to install and remove something (configuration can also be added here)
- `configure-` Used to configure something that is pre-installed on the ublue system (maybe shorten to `config-`?)
- `install-` Used by recipes where there is no functionality other than install
- `remove-` used by `remove-brew` due to the issues we got reported where it broke flatpak,  please ignore
- `toggle-` switch something on/off like for example `toggle-gnome-vrr` (if it makes sense maybe ditch `toggle-` and use `configure-` instead, unless we keep toggle for things that are specifically on/off switches?)
- `fix-` Apply a patch/workaround to something
- `thing` A single task that works as a shortcut and needs to be memorable, like `update` or `changelog`

Also i have added aliases to provide a shorter name for recipes too where i felt it was warranted.

Input about this proposed scheme is appreciated before merging

Snippet for choice dialogs (adjust for your needs)
```bash
#!/bin/bash
bold=$(tput bold)
normal=$(tput sgr0)
echo "${bold}Title${normal}"
echo 'text?'
OPTION=$(ugum choose option1 "option 2")
if [ "$OPTION" == "option1" ]; then
  echo 'stuff'
elif [ "$OPTION" == "option 2" ]; then
  echo 'other stuff'
fi
```

If you have a lot of options (more than 3?) you can do what i did for nix for readablilty
```bash
OPTION=$(ugum choose \
    option1 \
    "option 2" \
    "another option" \
    option4 \
  )
```
Using `ugum` (universal gum) will let it generate choice dialogs even when `gum` is not installed, they will look something like this:

With `gum` installed, gum will be prioritized:
image

Without `gum` but with `fzf` is installed, fallback to fzf:
image

Without both `fzf` and `gum`, fallback to pure bash:
image

### DIFF
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -9,7 +9,7 @@ bios:
 
 # Change the user's shell
 chsh new_shell:
-    #!/usr/bin/env bash
+    #!/usr/bin/bash
     set -euo pipefail
     if [ "{{ shell }}" = "{{ new_shell }}" ] ; then
       printf "Your shell is already set to %s.\n" "{{ new_shell }}"
@@ -24,7 +24,7 @@ chsh new_shell:
 
 # Regenerate GRUB config, useful in dual-boot scenarios where a second operating system isn't listed
 regenerate-grub:
-    #!/usr/bin/env bash
+    #!/usr/bin/bash
     if [ -d /sys/firmware/efi ]; then
       sudo grub2-mkconfig -o /etc/grub2-efi.cfg
     else
@@ -42,7 +42,7 @@ enroll-secure-boot-key:
 
 # Install or uninstall distrobox-git
 setup-distrobox-git:
-    #!/usr/bin/env bash
+    #!/usr/bin/bash
     echo 'Which version of Distrobox do you want to use?'
     OPTION=$(ugum choose "Latest distrobox-git snapshot" "uBlue provided version")
     if [ "$OPTION" == "Latest distrobox-git snapshot" ]; then
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -40,15 +40,21 @@ enroll-secure-boot-key:
     echo 'Enter password "ublue-os" if prompted after your user password.'
     sudo mokutil --import /etc/pki/akmods/certs/akmods-ublue.der
 
-# Upgrade Distrobox to the latest git version
-distrobox-git:
-    echo 'Installing latest git snapshot of Distrobox...'
-    curl -s https://raw.githubusercontent.com/89luca89/distrobox/main/install | sh -s -- --next --prefix ~/.local
-
-# Downgrades Distrobox to the Fedora version
-remove-distrobox-git:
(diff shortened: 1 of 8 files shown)


HikariKnight:
I have left `install-nixgl` as it is, since it has no other recipes like a removal recipe while with the `setup` verb one would expect install, remove and configuring something

HikariKnight:
Distrobox file will require some more thought on what would be the best approach, will do that one later as it is fine as it is for now.

HikariKnight:
> let's see what others think. I'd like to shove your reasoning for each verb in a readme at the root of the dir so we have it going forward, but we can do that in a follow up PR.

It is why i want to see what others think first before it gets fully committed to going forward

> Thanks for working on this!

Np! someone has to do it :smile:

EyeCantCU (approved):
The changes here look great! Definitely a significant improvement over what we have. Just a note - in the future, we should implement quick toggles/selections for these so that they can be used in other places like yafti. I.E., for 'configure-gamepads`, we could have `configure-gamepads stock` or `configure-gamepads ublue`

HikariKnight:
> The changes here look great! Definitely a significant improvement over what we have. Just a note - in the future, we should implement quick toggles/selections for these so that they can be used in other places like yafti. I.E., for `configure-gamepads`, we could have `configure-gamepads stock` or `configure-gamepads ublue`

need to play around with it and figure out how it works first in just and do some tests as i was thinking we could do something as simple as just feeding the ugum choice return value to the command or something.

EyeCantCU (approved):
Everything looks great. Bazzite just needs a couple things adjusted in yafti and then this will be good to go :)

noelmiller:
I agree with the above statement! I just added several more distroboxes in https://github.com/ublue-os/config/pull/185

We are also talking about wanting to support toolbox as well so users have a choice of which they prefer to use!

HikariKnight:
> I have a proposal:
> 
> Can we change the extremely long list of distrobox-`[image name here]` into one recipe called something like `install-distrobox`?. This recipe would have an optional arg to choose the image you want. If this arg is not entered then it would use ugum to let the user choose an option.
> 
> This would fit with the naming scheme you have created and would declutter the list of recipes when doing `ujust -l`

will be in separate PR as i have some ideas on how to improve it as i really want to shrink that list too, but i also want to enhance it by letting users apply a name and platform (docker/podman or toolbox) for the distroboxes.
But i will have to draft the change and check what would make sense. :smile:

KyleGospo (approved):

KyleGospo:
Did we fold the change to default to upstream gamepads and leave ours as the option into this?

bsherman:
I was just coming here to do a final read and approve! But it's in already!  

Thanks for the effort @HikariKnight

Issue #106, opened 2023-08-29
just: automate setting up RHEL containers


castrojo:
It should be easy for people who work with RHEL to use our stuff. Looks like what we need is documented here: 

https://debarshiray.wordpress.com/2023/08/25/fedora-meets-rhel-upgrading-ubi-to-rhel/

It'd be neat to set it up exactly how they want it with toolbx too, let's not make editorial decisions on their preferred tools.

bsherman:
Sounds like this is simply adding a `just` directive (or set of them) which would do this: `toolbox create --distro rhel --release 9.2` maybe with parameter... and ideally using distrobox....

But... maybe this would also be a reason to stop removing `toolbox`, too... RHEL users may really want it as a fallback for their use case.

noelmiller:
# My Head Hurts
So, the landscape is confusing in regards to Red Hat Toolbox Containers. I'm going to ask some questions tomorrow to hopefully get a better understanding of what customers and individuals should be using.

# UBI
Images are available to use completely for free and can be accessed from `registry.access.redhat.com` without a subscription. The FAQ clearly indicates that anyone can use these images. https://developers.redhat.com/articles/ubi-faq#

# quay.io RHEL toolbox images
Timothy has been managing these and they seem to be mostly up to date. They are slightly behind the official toolboxes which I will get to next. These exist here: https://quay.io/repository/toolbx-images/rhel-toolbox?tab=info

# Official Toolboxes
These require a subscription since they are hosted at `registry.redhat.io` This would require you to be a customer (or maybe a developer account can access these?) These seem to be the most up to date with current RHEL releases and are specifically designed to work inside of Toolbox. 

## RHEL 8 
https://catalog.redhat.com/software/containers/rhel8/toolbox/611bd18eeb34b73652492a75?architecture=amd64&image=6543c1d79fc22886ccc4df72&container-tabs=overview

## RHEL 9: 
https://catalog.redhat.com/software/containers/rhel9/toolbox/61532be3bc74e0be7ca66e40?architecture=amd64&image=6571634acb7b2ef3e5ae8937&container-tabs=overview

# Other Toolbox Images
From the offical [toolbox repo](https://github.com/containers/toolbox) there are several out of date containers there for RHEL: https://github.com/containers/toolbox/tree/main/images/rhel

# Conclusion
[...]

PR #185, opened 2024-01-09
feat(just): Add RHEL Toolboxes and UBI Images

noelmiller:
This is to fix: https://github.com/ublue-os/config/issues/106

I decided to go with the official toolbox images for RHEL which will require users to log in. I did add a check to confirm the user is logged in to `registry.redhat.io` before attempting to create the distrobox.

For the UBI images, I went with the URL that does not require authentication to pull them down.

### DIFF
--- a/build/ublue-os-just/30-distrobox.just
+++ b/build/ublue-os-just/30-distrobox.just
@@ -51,6 +51,36 @@ distrobox-centos:
     echo 'Creating Centos (Stream) distrobox ...'
     distrobox-create --nvidia --image quay.io/toolbx-images/centos-toolbox:latest -n centos -Y
 
+# Create a RHEL8 container 
+distrobox-rhel8:
+    #!/usr/bin/env bash
+    if ! grep -q "registry.redhat.io" $XDG_RUNTIME_DIR/containers/auth.json; then
+      echo "Please login using your Red Hat Account or Red Hat Developer Account ..."
+      podman login registry.redhat.io
+    fi
+    echo 'Creating RHEL 8 distrobox ...'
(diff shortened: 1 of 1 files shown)


noelmiller:
Giving users a choice definitely is better in my opinion (but I think we should default to distrobox). I'm not as familiar with Toolbox as I am with distrobox. 

That being said, it might make sense to have a separate just file specifically for toolbox since it is a different tool. Is it possible to import variable files into justfiles? That way, we could define all of our containers we want to use and make them available to both toolbox and distrobox.

bsherman:
> Giving users a choice definitely is better in my opinion (but I think we should default to distrobox). 
I agree, mostly from the position that our just recipes are all, already `distrobox-*` 

> That being said, it might make sense to have a separate just file specifically for toolbox since it is a different tool.
Per the newer restructured just recipe files, yes, I think that could make sense, but, also that seems like a lot of duplication instead of DRY.

>Is it possible to import variable files into justfiles? That way, we could define all of our containers we want to use and make them available to both toolbox and distrobox.

Yes. This latter thought is basically what I was suggesting in my second option... 

I'd be happy to merge this as is, but file a follow-up issue to be more "toolbox-friendly" and either coach you through the env var stuff or provide the PR myself.

noelmiller:
Follow up issue would be great! I'm happy to work together on this to get more familiar with `just`. I was just reading that you can use environment variables and export them. There are also parameters as well which we could explore!

bsherman (approved):
Per our discussion, I've created #186  and I approve these changes.

KyleGospo:
This looks great to me, don't even need to create the distrobox-* images now that distrobox overrides xdg-open. +1

KyleGospo (approved):

Issue #187, opened 2024-01-10
Enable Bash Completion for ujust


noelmiller:
`ujust` currently does not have bash completion which is annoying in comparison to using the standard `just` command. If the intention is to use `ujust` by default going forward, we should include this in our RPM build.

@HikariKnight has said he would like to work on this.

HikariKnight:
Merged, once new builds happen of ublue-os/config and new builds of all the images happen, ujust will have bash completion

PR #188, opened 2024-01-10
feat(just): Generate tab completion for ujust

HikariKnight:
Addresses #187 

Adds a postinstall step to the rpm spec file which will run
`just --completion bash | perl -pe 's/([\(_" ])just/\1ujust/g' > %{_datadir}/bash-completion/completions/ujust`

`just --completion bash` generates a just bash completion file and outputs it to STDOUT
`perl -pe 's/([\(_" ])just/\1ujust/g' > %{_datadir}/bash-completion/completions/ujust` rewrites the output using perl so that the bash completion will work for ujust and writes the output to an ujust bash completion file.

edit: replaced `perl -pe` with `sed -E` as we found out the regex worked in that and perl is not guaranteed to be on potential systems the rpm will be installed on

### DIFF
--- a/build/ublue-os-just/ublue-os-just.spec
+++ b/build/ublue-os-just/ublue-os-just.spec
@@ -59,7 +59,7 @@ install -Dm755 %{SOURCE10} %{buildroot}%{_bindir}/ugum
 
 %post
 # Generate ujust bash completion
-just --completion bash | perl -pe 's/([\(_" ])just/\1ujust/g' > %{_datadir}/bash-completion/completions/ujust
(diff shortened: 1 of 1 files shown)


bsherman:
For documentation purposes... I requested `sed -E` instead of `perl -pe` to avoid making perl a requirement of the just RPM.

bsherman (approved):
This looks good.

HikariKnight:
> For documentation purposes... I requested `sed -E` instead of `perl -pe` to avoid making perl a requirement of the just RPM.

which i agree with since the regex worked as expected in `sed -E`, perl is just a habit of mine whenever regex in scripts are involved :sweat_smile:

KyleGospo (approved):

PR #190, opened 2024-01-12
fix: change rpm-ostreed-automatic timer from 4AM local time to 4PM/16:00 UTC

Renner0E:
I think this makes slightly more sense than updating at 4AM/04:00 local time.
[Is this intentional or is this just a typo?](https://github.com/ublue-os/config/blob/3da21b60298c8b7a9a2ba745f134d7c2fd71e4b0/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf)

[The GitHub actions for main kick off at 3PM/15:00 UTC.](https://github.com/ublue-os/main/blob/cb54f2c85b07e6a51dcfc6e2fd38fd6eb6647de1/.github/workflows/build.yml)

Why not just check for updates as soon as they are available?
This would only benefit images that are finished building in GitHub before 4PM UTC.

Off the top of my head I can't think of any downsides compared to the current state.

### DIFF
--- a/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf
+++ b/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf
@@ -1,4 +1,4 @@
 [Timer]
 RandomizedDelaySec=10m
-OnCalendar=*-*-* 4:00:00
+OnCalendar=*-*-* 16:00:00 UTC
 Persistent=true


bsherman:
> I think this makes slightly more sense than updating at 4AM/04:00 local time. [Is this intentional or is this just a typo?](https://github.com/ublue-os/config/blob/3da21b60298c8b7a9a2ba745f134d7c2fd71e4b0/files/usr/etc/systemd/system/rpm-ostreed-automatic.timer.d/override.conf)

The current time is not meant to relate when builds occur in CI, etc. The idea is the machine will attempt to download updates during user's local timezone early morning, while idle, not middle of the day when resources are potentially more constrained. 

Also, if the machine is sleeping or off at this time, I it should attempt to run this after waking up because it "missed" the last chance.

Renner0E:
Alright thanks for the explanation.

Issue #186, opened 2024-01-09
provide option to use toolbox in our distrobox just recipes


bsherman:
I suggest maybe add an argument to all our distrobox recipes which allows toggling toolbox, and perhaps allows setting an env var which defaults to toolbox. The idea people that some users definitely do use it exclusively.

> Giving users a choice definitely is better in my opinion (but I think we should default to distrobox). 

I agree, mostly from the position that our just recipes are all, already `distrobox-*` 

>Is it possible to import variable files into justfiles? That way, we could define all of our containers we want to use and make them available to both toolbox and distrobox.

Yes. This latter thought is basically what I was suggesting in my second option... 

I'd be happy to merge [this](https://github.com/ublue-os/config/pull/185) as is, but file a follow-up issue to be more "toolbox-friendly" and either coach you through the env var stuff or provide the PR myself.

_Originally posted by @bsherman in https://github.com/ublue-os/config/issues/185#issuecomment-1883388950_

bsherman:
Relates to #106

HikariKnight:
when i get to reorging the distrobox just file, i want to make it possible to select image, set a name (using a just arg) and platform (distrobox or toolbox) as part of the overhaul, but since this would require some more work than the rest of the reorg, i decided to split reorging the distrobox just file as its own PR afterwards. 😃

bsherman:
@HikariKnight this relates to your comment here: https://github.com/ublue-os/config/pull/181#issuecomment-1884266421

Even though you are on the discussion already, linking it for reference.

HikariKnight:
Just updating that i am laying the groundwork for this now in https://github.com/ublue-os/config/pull/194
However i am not sure if everything that has a docker image has a toolbox image and UBI image so it would be nice to figure out the logic before putting it together.

bsherman:
It was discussed and determined that neither #194 nor #199 either addressed nor hindered this issue.

Also, an important point was raised, this issue is really about enabling `toolbox` for rhel and basic fedora toolboxes, not generically attempting to support all our various distroboxes with `toolbox` as that can't work given different feature sets.

HikariKnight:
Should we try make the just recipe work similar to `ujust distrobox-assemble` where it reads from a simplified .ini manifest and generates a list of images from that.
I am thinking something like
```ini
[rhel-ubi8]
image=registry.access.redhat.com/ubi8/toolbox:latest
distro=
release=

[rhel9]
image=
distro=rhel
release=9
```

[...]

PR #194, opened 2024-01-13
feat(just): Add sourceable tooling for just and refine reorg

HikariKnight:
The idea of this PR is to add a file that our just recipes can source which will import useful functionality like text formatting, colors and UI tools (instead of having to use the same ugum commands over and over)
image
image

image
NOTE: using the fallback bash version for demonstration purposes

Also please note that function names are not decided yet, these were chosen for demonstration purposes only.

This is very WIP and barebones, so i would like to gather any functionality, colors, formatting, etc. you would like added in here that would make sense.

if you are unsure about the bash color code for said color just give me the hex code for it and i will find the closest one available. :smile: 

Also where should we keep the source file?
`/usr/libexec/ujust/`
`/usr/share/ublue-os/justlib/`
other location that might make more sense?


HikariKnight:
Reminder to myself: look into making a distrobox assemble function that is portable

from myself on discord:

> i think right now to avoid complexity, there will be 2 distrobox just recipes in config
> `distrobox-new image name [homedir] [extra_args]`
> `alias distrobox=distrobox-new`
> `distrobox-import [file] [extra_args]`
> 
> import will list all distroboxes offered in the file and let the user select which one to assemble or they can just supply an arg to do all, etc
> default file will be /etc/distrobox/distrobox.ini

NOTE: this is just me brainstorming

HikariKnight:
Lots of changes, note for @castrojo, bluefin needs https://github.com/ublue-os/bluefin/pull/817 merged after config is built

HikariKnight:
I added the `distrobox.ini` file to `files/usr/etc/distrobox/` and the scripts use `/usr/etc/distrobox/distrobox.ini`
Looking at the containerfile it looks like i did not need to update that, but if i saw wrong please correct me

bsherman (changes requested):
Wowsas... there's a lot here.

So, this PR seems to be at least PRs combined in one.

First, there's the stuff that directly applies to #186 

Second, there's a bunch of stuff related to just file abstraction, theming, etc.

I'd really like to split this into distinct PRs because the first issue is a pre-existing issue which was created to solve a usability problem, and the second seems to be more of a developer refactor. May be valid, but it's better if we can grok them independently and give feedback to each respective goal.

Since it seems like you felt the refactor was helpful for the goal of cleaning up distrobox/toolbox usability... I'd suggest removing the distrobox/toolbox changes and focusing on the refactor first.

HikariKnight:
> Wowsas... there's a lot here.
> 
> So, this PR seems to be at least PRs combined in one.
> 
> First, there's the stuff that directly applies to #186
> 
> Second, there's a bunch of stuff related to just file abstraction, theming, etc.
> 
> I'd really like to split this into distinct PRs because the first issue is a pre-existing issue which was created to solve a usability problem, and the second seems to be more of a developer refactor. May be valid, but it's better if we can grok them independently and give feedback to each respective goal.
> 
> Since it seems like you felt the refactor was helpful for the goal of cleaning up distrobox/toolbox usability... I'd suggest removing the distrobox/toolbox changes and focusing on the refactor first.

I will split them tomorrow after work then, it is late here

HikariKnight:
> > I will split them tomorrow after work then, it is late here
> 
> Thanks for understanding and all the effort! No rush from me.

i decided it isnt too much to get done before going to bed so splitting them now

build/ublue-os-just/ublue-os-just.spec
```diff
@@ -1,7 +1,7 @@
 Name:           ublue-os-just
 Packager:       ublue-os
 Vendor:         ublue-os
-Version:        0.20
+Version:        0.22
```
bsherman, line 4: Small thing but, I'm curious why this jumped from 0.20 to 0.22? Where is 0.21?

bsherman: Oh. I answered myself. Looks like we had a 0.21 changelog entry which missed the package version update.

HikariKnight: yep so i fixed it when i noticed it

build/ublue-os-just/libjust.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/bash
+# shellcheck source=/dev/null
+# Import color formatting
+source /usr/lib/ujust/libcolors.sh
+# Import text formatting
+source /usr/lib/ujust/libformatting.sh
+# Import functionality for just
+source /usr/lib/ujust/libfunctions.sh
```
bsherman, line 8: Minor suggestion for future maintenance and improved organization of the bash lib files.

What made me start thinking about this: it's slightly inconsistent that we have `/usr/lib/ujust` but the main file is `libjust.sh` not `libujust.sh` or `ujust.sh`.

So one suggestion...

Within the source `ublue-os-just`:
- make a subdir `lib-ujust/`
- move all the `lib*.sh` files into the subdir, which keeps them clearly separated from other scripts which may end up in that source dir
- rename `libjust.sh` to `ujust.sh`
- the end result for `/usr/lib/ujust` on a users filesystem will be the same as this code already will install, but with `/usr/lib/ujust/ujust.sh` as a clear entrypoint which includes the other `lib*.sh` files there.

build/ublue-os-just/libfunctions.sh
```diff
@@ -0,0 +1,167 @@
...
+        # shellcheck disable=SC2086
+        CHOICE=$(Choose ALL $CONTAINERS)
+    fi
+
+    # If choice is not empty by now (will be empty if escaped from Choice function)
+    if [ -n "$CHOICE" ]; then
+        # Assemble the selected container
+        Assemble "$ACTION" "$FILE" "$CHOICE"
+    fi
+}
```
bsherman, line 167: Seems like the `Distrobox`, `Assemble`, and `AssembleList` functions are all unused in this PR.

Looks like they are meant to be be part of #197 

Can we move them to that PR please?

build/ublue-os-just/libjust.sh
HikariKnight: this is what feedback is good for since i was not entirely sure myself what to name it so i went with `libjust` (or `justlib` for a bit) so that i had something to at least test with until things got ready. i will however move the files and change `libjust.sh` to `ujust.sh` so that is done before merge.

bsherman: If I missed some other inconsistency, I'd approach it the same way... You've built `libujust` which will live at `/usr/lib/ujust` ... I think the `ujust` distinction is important since it's heavily tied to the `ujust` implementation and won't impact standard `just` usages (unless a user very specifically searched it out and imported it in their Justfile`.

build/ublue-os-just/libfunctions.sh
HikariKnight: will do

HikariKnight: think i did it correctly, should be in #197 now

build/ublue-os-just/ublue-os-just.spec
```diff
@@ -49,20 +53,32 @@ mkdir -p -m0755  %{buildroot}%{_bindir}
 install -Dm755 %{SOURCE9} %{buildroot}%{_bindir}/ujust
 install -Dm755 %{SOURCE10} %{buildroot}%{_bindir}/ugum
 
+# Add bash library for use in just
+mkdir -p -m0755 %{buildroot}%{_libdir}/ujust/
+install -Dm644 %{SOURCE12} %{buildroot}%{_libdir}/ujust
```
bsherman, line 58: Looks like `${_libdir}` varies by platform...  so I'm thinking maybe this didn't get tested. ;-)

My suggestion on testing is:
1. build the config container locally to some image like `ubconfig`
2. add the rpms from the config layer in a simple container built FROM a fedora toolbox image
3. install the RPM and test `ujust` configs, etc

Doing the above, I saw the following:
```
[root@e0cd6e40ecbd share]# ujust|head -n2
/tmp/justzB7Bbq/_default: line 6: /usr/lib/ujust/libformatting.sh: No such file or directory
/tmp/justzB7Bbq/_default: line 7: Urllink: command not found

Available commands:
[root@e0cd6e40ecbd share]# rpm -ql ublue-os-just 
/etc/profile.d/ublue-os-just.sh
/usr/bin/ugum
/usr/bin/ujust
/usr/lib64/ujust/libcolors.sh
/usr/lib64/ujust/libformatting.sh
/usr/lib64/ujust/libfunctions.sh
/usr/lib64/ujust/ujust.sh
/usr/share/ublue-os/just
/usr/share/ublue-os/just/00-default.just
/usr/share/ublue-os/just/10-update.just
/usr/share/ublue-os/just/20-clean.just
/usr/share/ublue-os/just/30-distrobox.just
/usr/share/ublue-os/just/40-nvidia.just
/usr/share/ublue-os/just/50-akmods.just
/usr/share/ublue-os/just/60-custom.just
/usr/share/ublue-os/just/70-nix.just
/usr/share/ublue-os/justfile
```

Suggestions on how to fix are either be smarter about the path in the source side (maybe complicated) or be more restrictive in the spec file `install`... I'd probably choose the latter.

I *think* that would change the target path for ujust lib stuff:
from: `%{buildroot}%{_libdir}/ujust`
to: `%{buildroot}/usr/lib/ujust` 

Please try it and see if you can reproduce what I saw and if the suggestion then fixes it.

Thanks!

bsherman (changes requested):

build/ublue-os-just/ublue-os-just.spec
HikariKnight: Ah yeah originally i used `/usr/libexec` and when i decided lib would be better, i looked up the macro for it in the fedora docs and did not notice it led to `/usr/%{_lib}`
and i agree being more restrictive in the spec file is the way to go.

build/ublue-os-just/ublue-os-just.spec
```diff
@@ -49,20 +53,31 @@ mkdir -p -m0755  %{buildroot}%{_bindir}
 install -Dm755 %{SOURCE9} %{buildroot}%{_bindir}/ujust
 install -Dm755 %{SOURCE10} %{buildroot}%{_bindir}/ugum
 
+# Add bash library for use in just
+mkdir -p -m0755 %{buildroot}/usr/lib/ujust/
+install -Dm644 %{SOURCE12} %{buildroot}/usr/lib/ujust
+install -Dm644 %{SOURCE13} %{buildroot}/usr/lib/ujust
+install -Dm644 %{SOURCE14} %{buildroot}/usr/lib/ujust
+install -Dm644 %{SOURCE15} %{buildroot}/usr/lib/ujust
```
bsherman, line 61: Apologies, I know that I previously suggested `%{buildroot}/usr/lib/ujust` but I realized hard-coding `/usr` breaks the concept of exec-prefix. And while it probably would work in this situation, it's not ideal, and it's nice when we can do things "right".

I found a helpful reference which confirmed my minor concern: https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/

So, without more information about "best practice",  I think I'm indifferent on if we use:
- `%{buildroot}/%{_exec_prefix}/lib/ujust`
-  or `%{buildroot}/%{_datadir}/ujust`

 but i do think either of these is better than fully hard-coding `/usr/lib/ujust` in the spec file.

KyleGospo (approved):

PR #199, opened 2024-01-18
feat(just): add distrobox tooling to just and cleanup distrobox just file

HikariKnight:
split from #194

### DIFF
--- a/build/ublue-os-just/30-distrobox.just
+++ b/build/ublue-os-just/30-distrobox.just
@@ -1,93 +1,107 @@
 # vim: set ft=make :
 
-# Create an Arch container
-distrobox-arch:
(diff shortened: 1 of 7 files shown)


KyleGospo:
Bazzite will need changes based on this before it lands

EyeCantCU (approved):
This looks good to me :). DNM until Bazzite is up to speed

HikariKnight:
> Bazzite will need changes based on this before it lands, a few of our just commands create a fedora distrobox and action upon it

Changes here should not affect bazzite as the container made for applications is the same one used in the old commands
but i will get a PR ready

KyleGospo (approved):

PR #200, opened 2024-01-20
fix(just): move default distrobox manifests into ujust rpm

HikariKnight:
When i made the reorg of the distrobox just files i was told by someone i could just add the ini files in their path under `files` and they would be added to the image, turns out that was not the case so i have moved them into the ujust rpm (which after thinking about it is a better idea since they are used by ujust recipes) so users can get working distrobox commands.


EyeCantCU (approved):
LGTM :)

p5 (approved):

PR #203, opened 2024-01-24
chore: Drop gamepad configuration script thanks to changes in ublue-os/akmods no longer requiring it

KyleGospo:
See: https://github.com/ublue-os/akmods/pull/115


bsherman (approved):
With us no longer including a driver that breaks xpad, there's no need to toggle... the new xone(without xpad support) just works as an addon.

EyeCantCU (approved):

PR #207, opened 2024-01-28
added installation of AM manager and removal to 00-default.just

dnkmmr69420:
https://github.com/ivan-hc/AM-Application-Manager#uninstall


castrojo:
This should be in a custom image, not in ublue itself.

PR #212, opened 2024-01-31
feat: add udev rules for linksys ae1200 usb dongle

another-sapiens:
Provides support for the Linksys AE1200, a USB dongle that enables WiFi.

Essentially, on normal distros the dongle does not work on boot. It must be physically disconnected and plugged back in to be recognized. Upon closer examination, there seems to be a mismatch between the device ID on boot (`13b1:0bdc`) and the value that `brcmfmac` (the kernel module) accepts (`13b1:0039`). After physically plugging it for the 2nd time, `lsusb` states the correct ID.  

Having this udev rule fixes the issue and WiFi works on boot. Merging this would remove the need to keep the spare `.rules` file.


HikariKnight:
We use Semantic PR commit messages as outlined here
https://universal-blue.org/CONTRIBUTING/#commit-messages

other than that the changes look good

another-sapiens:
Hello. I don't know if there's a cleaner way, so I will close this PR and create another one.

Issue #210, opened 2024-01-30
Toolbox Distrobox Incompatabilities


noelmiller:
When a distrobox is created, it shows up in the list of Toolbox Containers that you can use. This is an issue because you receive the message below:
```
nomiller@mirkwood:~$ toolbox enter alma 
Error: container alma is too old and no longer supported 
Recreate it with Toolbox version 0.0.17 or newer.
```

We need to find a solution either in Distrobox or Toolbox and contribute that way. Otherwise, we can patch our toolbox RPM to fix this issue if we cannot find a solution upstream.

bsherman:
I'm tinkering with some way to patch `toolbox` such that it filters out "toolbox" containers which were not created by it.

I think it's too specific for toolbox to have awareness of distrobox, but there are other ways to filter.

m2Giles:
It's looking for the label "com.github.containers.toolbox"

If it's set at all, even to none. It will show up in the toolbox list. The only way around this was not building the container with that label.

Distrobox on the other hand is looking for the following:

distrobox in the mounts section
distrobox in the labels section

Distrobox will see our toolboxes since we have distrobox in the label usage even when used as quadlet with no mounts.

I tried removing both and now the container will show up in prompt as a regular podman container. However, it won't use the same scarf and will not properly enter if it cannot determine if it's a toolbox or distrobox.

bsherman:
> The only way around this was not building the container with that label.

Clarification, not building the IMAGE with the label.

bsherman:
This isn't actually completed as the problem described still exists. 

With consensus, I think we could close as "won't fix" since it's really something for toolbox/distrobox to sort out, not us.

HikariKnight:
i agree, it is not really our job to fix this

PR #218, opened 2024-02-03
feat(just): add option to use toolbox instead of distrobox

HikariKnight:
Since `toolbox` and `distrobox` are nowhere near feature parity ([and not always compatible with each other](https://github.com/ublue-os/config/issues/210)), it made more sense to provide the toolbox option as a separate recipe. 

## Recipes
This PR implements the recipes `toolbox-new` and a **very** limited version of `distrobox-assemble` in the form of `toolbox-assemble` for ujust.
There are no plans to expand upon this, the assemble feature is provided as a nice to have for defining multiple toolbox containers and assembling them with 1 command.

## Internals
For internal ujust functions this implements `Toolbox`, `ToolboxAssemble` and `ToolboxAssembleList` that will work similar to the internal ujust functions for distrobox.

### Why make toolbox internal functions?
The easiest maintainable way to generate a list of images is through a similar `ini` manifest file like we do for distrobox, we already had the logic for it, so why not re-use it and avoid having to update the script functions/just file whenever the images supplied gets too old.

image
NOTE: Dev container does not have gum or fzf installed, the bash fallback is perfect for demonstration images like this

This should suffice as an implementation for #186


HikariKnight:
This is now ready for review

KyleGospo (approved):

m2Giles (approved):

PR #221, opened 2024-02-16
fix(just): enable nix units after install

HikariKnight:
Fixes nix install as the new installer does not seem to handle this by itself.
Ice from discord is the one that provided the fix


HikariKnight:
Tested and works when tested locally

KyleGospo (approved):

p5 (approved):

PR #235, opened 2024-03-24
feat: initial brew installer

gerblesh:
This PR basically copies the brew installer logic from bluefin:
https://github.com/ublue-os/bluefin/blob/main/usr/etc/profile.d/brew.sh
https://github.com/ublue-os/bluefin/blob/main/Containerfile#L104

only change is using `curl https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh | /usr/bin/bash -c` instead of having the install script directly on the image, might be better to directly pull these scripts down when building to keep consistency

### DIFF
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -63,6 +63,30 @@ setup-distrobox-git:
       exit 0
     fi
 
+# installs homebrew on the system
+install-brew:
+    #!/usr/bin/env bash
+    source /usr/lib/ujust/ujust.sh
+    if [[ ! -f "/var/home/linuxbrew/.linuxbrew/bin" || ! -x "/var/home/linuxbrew/.linuxbrew/bin/brew" ]]; then
+        echo "${b}Brew Installation${n}"
+        echo "Please ${b}IGNORE${n} everything the installer tells you to do at the end"
+        echo "We have already done it for you! You just need to close and re-open the terminal after installation"
+        echo "Do you understand?"
+        echo "Please type in \"YES I UNDERSTAND\" and press enter"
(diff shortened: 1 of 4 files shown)


gerblesh:
Brew install/uninstall targets seem to work from my testing. One issue is that the `install-brew` target conflicts with the bluefin `install-brew` so changes might need to be made in upstream bluefin for this to be merged

bsherman (commented):
Why shouldn't we include this file in this PR?

https://github.com/ublue-os/bluefin/blob/main/usr/etc/profile.d/brew.sh

bsherman (changes requested):
Please add `/usr/etc/profile.d/brew.sh` to the SPEC file.

bsherman:
RPM fails to build:
```
+ install -Dm755 /tmp/ublue-os/rpmbuild/SOURCES/brew.sh /tmp/ublue-os/rpmbuild/BUILDROOT/ublue-os-just-0.30-1.fc39.x86_64/etc/profile.d/brew.sh
install: cannot stat '/tmp/ublue-os/rpmbuild/SOURCES/brew.sh': No such file or directory
```

bsherman (approved):
This looks good! Tested locally and seems to do what we expect!

Should probably file an issue in https://github.com/ublue-os/bluefin and/or PR to make changes given the conflict there after this merges, but good stuff.

castrojo (approved):

citrixscu:
@bsherman do we also need to add `[dx]/usr/share/fish/vendor_conf.d/brew.fish` to SPEC? Things seem to work fine with bash default shell, but not sure what happens if fish is set to the default login shell.

PR #242, opened 2024-04-02
feat: added Fido2/Yubi Key Setup with ujust

thedragonkeeper:
This addition creates a ujust script that allows easy config of a fido/yubi key on to an encrypted disk.
it provides options to:
- enrol a new key
- manually assign device and luks volume if you have multiple encrypted disks
- set the key has the default method.   This allows the key to be the only auth method needed and password as fallback.  Great for steamdeck use if fido key doesnt need a passphrase as it doesnt yet have a onscreen keyboard for this
- rebuilds initramfs with fido module - its an option incase you want to do thjis for multiple volumes, it doesnt need rebuilding again.
and it updates /etc/crypttab with fido option for selected volume if needed

### DIFF
--- a/build/ublue-os-just/50-akmods.just
+++ b/build/ublue-os-just/50-akmods.just
@@ -29,7 +29,7 @@ configure-broadcom-wl ACTION="prompt":
       echo "${bold}Disabled${normal} Broadcom Wireless, please reboot for changes to take effect"
     fi
 
-# Configure/Add Fido/Yubi key to an Encrypted disk with key priority 
+# Configure/Add Fido/Yubi key to an Encrypted disk with key priority
 configure-fido2:
     #!/usr/bin/bash
     if [ "$EUID" -ne 0 ]
@@ -93,7 +93,6 @@ configure-fido2:
         [Yy]* ) systemd-cryptenroll --fido2-device=auto --fido2-with-user-verification=true $device;;
         *) echo "not enrolling a new key" ;;
     esac
-    ### edit the crypttab like for the luks volume if needed
     luksline=$(grep -r $luks /etc/crypttab)
     if [[ ! "$luksline" = *"fido2-device"* ]] ; then
         opts=$(echo $luksline | awk '{print $4}')
--- a/build/ublue-os-just/50-akmods.just
+++ b/build/ublue-os-just/50-akmods.just
@@ -28,3 +28,57 @@ configure-broadcom-wl ACTION="prompt":
       sudo bash -c 'echo "blacklist wl" > /etc/modules-load.d/default-disable-broadcom-wl.conf'
       echo "${bold}Disabled${normal} Broadcom Wireless, please reboot for changes to take effect"
     fi
+
+# Configure/Add Fido/Yubi key to an Encrypted disk with key priority 
+fido2:
+    #!/usr/bin/bash
+    if [ "$EUID" -ne 0 ]
+        then echo "Please run as root"
+        exit
+    fi
+    clear; lsblk
+    ## find default devices
+    device='/dev/'$(lsblk -r --fs | grep crypto | awk '{print $1}')
+    luks=$(lsblk -r --fs | grep luks | awk '{print $1}')
+    echo " _______________________"
+    echo Your device is $device
+    echo Your crypto volume is $luks
+    # add custom values if needed
+    read -p "are these values correct? y/n" yn
+    case $yn in
+        [Nn]* ) read -p "enter the device volume? " device ; read -p "Enter value for luks volume " luks;;
+    esac
+    ## make sure custom value has /dev added
+    [[ $device = */dev/* ]] || device=/dev/$device
+    clear; lsblk
+    echo Your device is $device
+    echo Your crypto volume is $luks
+    ## enroll a key
+    read -p "do you need to enrol a new key?  if yes insert key before answering  y/n " yn
+    case $yn in
+        [Yy]* ) systemd-cryptenroll --fido2-device=auto --fido2-with-user-verification=true $device;;
+        *) echo "not enrolling a new key" ;;
+    esac
+    ### edit the crypttab like for the luks volume if needed
+    luksline=$(grep -r $luks /etc/crypttab)
+    if [[ ! "$luksline" = *"fido2-device"* ]] ; then
+        opts=$(echo $luksline | awk '{print $4}')
+        newopts=$(echo $luksline | sed "s/$opts/fido2-device=auto,$opts/g")
+        sed -i "s/$luksline/$newopts/g" /etc/crypttab
+        echo "updated crypttab with fido"
+    fi
+    ### sets the password to not be used unless a key is not present
+    read -p "Set the new key as the sufficient auth method? y/n " yn
+    case $yn in
+        [Yy]* ) cryptsetup config --key-slot 0 --priority ignore $device ; cryptsetup config --key-slot 1 --priority prefer $device;;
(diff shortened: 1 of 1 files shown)


HikariKnight:
some problems here
* We use Semantic PR so commit messages should follow: https://www.conventionalcommits.org/en/v1.0.0/
* Recipe should be named `configure-fido2` following the current naming convention used throughout ujust
* Your current implementation uses `rpm-ostree initramfs` and you will wipe out any initramfs arguments that has previously been set on the system (be it nvidia modules or other hwe related initramfs stuff), you should really consider putting it into a `dracut.conf` file inside `/etc/dracut.conf.d`

build/ublue-os-just/50-akmods.just
```diff
@@ -28,3 +28,57 @@ configure-broadcom-wl ACTION="prompt":
       sudo bash -c 'echo "blacklist wl" > /etc/modules-load.d/default-disable-broadcom-wl.conf'
       echo "${bold}Disabled${normal} Broadcom Wireless, please reboot for changes to take effect"
     fi
+
+# Configure/Add Fido/Yubi key to an Encrypted disk with key priority 
+fido2:
```
HikariKnight, line 33: should be `configure-fido2` so that if follows current recipe naming verbs

build/ublue-os-just/50-akmods.just
```diff
@@ -28,3 +28,121 @@ configure-broadcom-wl ACTION="prompt":
       sudo bash -c 'echo "blacklist wl" > /etc/modules-load.d/default-disable-broadcom-wl.conf'
       echo "${bold}Disabled${normal} Broadcom Wireless, please reboot for changes to take effect"
     fi
+
+# Configure/Add Fido/Yubi key to an Encrypted disk with key priority
```
HikariKnight, line 32: just syntax checker fails here due to whitespace at the end

build/ublue-os-just/50-akmods.just
```diff
@@ -28,3 +28,121 @@ configure-broadcom-wl ACTION="prompt":
...
+        [0] ) cryptsetup config --key-slot 0 --priority ignore $device ; cryptsetup config --key-slot 1 --priority prefer $device;;
+        [1] ) cryptsetup config --key-slot 0 --priority normal $device ; cryptsetup config --key-slot 1 --priority prefer $device;;
+        *) echo "invalid option,  leaving slot priorities the same" ;;
+    esac
+    echo "
+
+    "
+    read -p "Enable the initramfs to rebuild ?  y/n " choice
+    case $choice in
+        [Yy]* ) rpm-ostree initramfs --enable;;
```
HikariKnight, line 145: unsure if this is needed if dracut modules are already enabled, but will have to double check if main images have dracut modules enabled by default. (will be able to check that tomorrow)

if they are not enabled or **if you need to** issue a rebuild then what you will have to do (thanks to how `rpm-ostree initramfs` functions) get the current initramfs arguments (with quotes in the right places for each `--arg=`) from `rpm-ostree -b | grep Initramfs | sed 's/Initramfs: //'` and rebuild the previously used `rpm-ostree initramfs` command to avoid wiping out the current args.

TheDragonkeeper: it seems that running this at this point makes it rebuild anyway, but if you have a better method im happy to change it. from my testing (on a bazzite build) a rebuild was needed for this module to work as intended.

and it seems that, according to rpm-ostree,    --arg  appends the line to the dracut arguments   and --enable Enable regenerating initramfs locally using dracut

it shouldnt be wiping anything out when doing a rebuild afaik  each deployment seems to retain the initramfs args from the last

m2Giles: Only Bazzite enables initramfs by default.

Additionally unsure if the force_drivers is necessary for fido. The dracutmodule should suffice for pulling it in. At least it does that for tpm2.

Should we have these security devices just specified in a file in /usr/lib/dracut.conf.d. if they are already present, we would only have to enable initramfs generation, do the key enrollment, and specify the fido device as a karg.

Key enrollment would be the only item that needs elevated privileges.

HikariKnight: > it seems that running this at this point makes it rebuild anyway, but if you have a better method im happy to change it. from my testing (on a bazzite build) a rebuild was needed for this module to work as intended.
> 
> and it seems that, according to rpm-ostree, --arg appends the line to the dracut arguments and --enable Enable regenerating initramfs locally using dracut
> 
> it shouldnt be wiping anything out when doing a rebuild afaik each deployment seems to retain the initramfs args from the last

from experience if you happen to lets say have `--arg="-I /etc/crypttab /etc/modprobe.d/nvidia.conf"` as the initramfs arguments and you run `rpm-ostree initramfs --enable`, `rpm-ostree initramfs` in its infinite wisdom will remove `--arg="-I /etc/crypttab /etc/modprobe.d/nvidia.conf"` as it was never supplied and the nvidia.conf modprobe never gets added and the nvidia driver does not get loaded. we had to work around this problem in bazzite which is why we have our own rebuilder script that gets triggered by `sudo touch /etc/bazzite/initramfs/rebuild` however it is currently not compatible with other images.

HikariKnight: > Only Bazzite enables initramfs by default.
> 
> Additionally unsure if the force_drivers is necessary for fido. The dracutmodule should suffice for pulling it in. At least it does that for tpm2.
> 
> Should we have these security devices just specified in a file in /usr/lib/dracut.conf.d. if they are already present, we would only have to enable initramfs generation, do the key enrollment, and specify the fido device as a karg.
> 
> Key enrollment would be the only item that needs elevated privileges.

i like the sound of this idea, i should also just look into the idea of a `initramfs-rebuild` that issues a rebuild, nothing else specifically for situations like this. However ideally everything would use initramfs-etc but that would require us to get the other bits sorted out first.

m2Giles:
Since this has been written, Bazzite and Bluefin/Aurora have started shipping an initramfs built as part of the build pipeline.

This reduces the need for the initramfs components, but is still needed on our main images.

HikariKnight (approved):
With the new initramfs changes done, i think this looks good enough now, sorry for the delay @TheDragonkeeper

storopoli:
This can be rebased and use the new tips that were merged in #266.

storopoli:
> * set the key has the default method.   This allows the key to be the only auth method needed and password as fallback.

Does this means that even if I set a PIN in my FIDO key, it will bypass it and unlock LUKS without the need to insert the PIN? (dangerous 😱)

thedragonkeeper:
> > * set the key has the default method.   This allows the key to be the only auth method needed and password as fallback.
> 
> Does this means that even if I set a PIN in my FIDO key, it will bypass it and unlock LUKS without the need to insert the PIN? (dangerous 😱)

this means that the password for LUKS doesnt need to be used. 
as you can set the priortiy or even disable it.
this isnt anything to do with the pin for the fido. 

i dont use a pin on my fido, the reason being that i only use this on a steam deck and that currently doesnt have a good way to input text,  it would need an onscreen keyboard added, so if it asked me for a pin id need to plug in a keyboard, which isnt practical.

bsherman:
I believe as is, this Connor be merged. And even the last comments were still asking for some changes. 

Plus, it's tightly coupled to encrypted LUKS volumes, so if desired should probably be resubmitted in the context of the ublue-os-luks package. 

That package source has moved to the packages repo if you'd like to submit again. 

Thank you for the effort and patience.

Issue #246, opened 2024-04-07
Feature Request/Proposal: Set up preferred shell


Sharkitty:
## Description
Adding a script usable by either `just` and/or `yafti` that should be able to let users choose their preferred shell, and set it up.

The minimum implementation I would need is setting up `zsh` and `omz` in this way:
- Checking if `zsh` is installed. If it's not, proposing the user to install `zsh` as a layered package.
- Propose the user to run `ohmyzsh` install script.

But I think it would make sense to extend this idea so users could choose any shell they want and the script would set it up.

I can work on this script, at least as far as the `zsh` part go. For further contribution it depends on the time I have.

castrojo:
We have `ujust chsh zsh` already. I think this should probably be `ujust oh-my-zsh` or perhaps an option step after the `ujust chsh zsh` to toggle oh-my-zsh on and off.

Sharkitty:
> We have `ujust chsh zsh` already. I think this should probably be `ujust oh-my-zsh` or perhaps an option step after the `ujust chsh zsh` to toggle oh-my-zsh on and off.

I missed this script ^^' But it just sets the login shell. So it could be a separate script called `ujust omz` or `ujust ohmyzsh` (or both). But I could also make something like `ujust chsh zsh --ohmyzsh` and if it's not specified it asks the user "do you want to use ohmyzsh framework?"

Shouldn't it be a script that `ujust` just calls so it can be used by `yafti` as well?

Agreed on making it a toggle however!

castrojo:
Yeah a standalone script in `/usr/libexec` would do the trick, that way we can call it from ujust and from yafti.

noelmiller:
The only response I have is I'm not sure this would belong in config? I can see something like this being beneficial in Bluefin DX, but not so much in Bazzite or other downstreams not focused on development.

castrojo:
I've noticed a large amount of shell diversity in our community.

I think letting each shell subcommunity hook up their own little config setups and stuff is fine -- people dig oh-my-zsh, I don't think it hurts to have it in config.

If people decide otherwise then I'll 100% take it in Bluefin!

bsherman:
I would agree that OMZ is a good fit for aurora/bluefin/bazzite ... but this my general concern with `config` echoes @noelmiller's above... 

In general, I don't think we have a clear guide to when it is appropriate to add things for those downstreams to the `config` since it's also pre-installed on main/hwe images.

castrojo:
[...]

PR #249, opened 2024-04-10
feat: Toggle omz script

Sharkitty:
## Description
Adding a script to install/uninstall `ohmyzsh` easily, and adding `just` recipes for it. The reason the implementation is in a separate script and not directly in the just files, is to make this script usable in `yafti` as well.

## Context
This feature is related to #246. I've already tested this downstream and I'm proposing this upstream because @castrojo showed interest in adding this feature to Bluefin as well.

### shell script
The shell script can either be called directly, without argument, in which case it will check if `omz` is already installed, if it is it will propose the user to uninstall it. Otherwise it will propose the user to install it.

You can directly call the `install` or `uninstall` subcommands as arguments.

These flags are also available:
- `-i` / `--interactive` -> Use this script interactively (Default).
- `-n` / `--non-interactive` -> Use this script non interactively (This is especially useful if you want to call this from a GUI.
- `-h` / `--help` -> Print this help page and exit.

### just recipes
There are three `just` recipes available:
- `ujust toggle-ohmyzsh` / `ujust toggle-omz` -> Install `ohmyzsh` if it's not installed, uninstall it if it's already installed.
- `ujust install-ohmyzsh` / `ujust install-omz` -> Install `ohmyzsh`.
- `ujust uninstall-ohmyzsh` / `ujust-uninstall-omz` -> Uninstall `ohmyzsh`.

## Additional note
You might need to logout/log back in or restart when `chsh` for the change to work correctly.

## Testing
- Run `ujust toggle-omz` twice.
- Run `ujust install-omz`.
- Run `ujust uninstall-omz`.
- You can also play around with the shell script directly.

### DIFF
--- a/build/ublue-os-just/01-omz.just
+++ b/build/ublue-os-just/01-omz.just
@@ -1,0 +1,17 @@
+alias toggle-omz := toggle-ohmyzsh
+
+toggle-ohmyzsh:
+	#!/bin/bash
+	/usr/libexec/toggle_omz.sh
+
+alias install-omz := install-ohmyzsh
+
+install-ohmyzsh:
+	#!/bin/bash
+	/usr/libexec/toggle_omz.sh install
+
+alias uninstall-omz := uninstall-ohmyzsh
+
+uninstall-ohmyzsh:
+	#!/bin/bash
+	/usr/libexec/toggle_omz.sh uninstall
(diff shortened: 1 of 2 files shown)


bsherman:
I would agree that OMZ is a good fit for aurora/bluefin/bazzite ... but this my general concern with `config` echoes @noelmiller's above... 

when is appropriate to add things for those downstreams to the `config` which is also pre-installed on main/hwe images?

bsherman:
Closing as the original issue is closed #246

PR #274, opened 2024-05-12
fix(just): adjust clean-system script to not delete turned off distrobox containers

wolfyreload:
* Replaced `podman system prune -a` with a safer command to just remove podman volumes and images
* Updated script comment


castrojo (approved):
Nice catch!

m2Giles (approved):

PR #278, opened 2024-05-16
chore: split brew from default.just

m2Giles:
Both Bazzite and Bluefin/Aurora are adding brew on image. Because we will not want people to use the ujust commands anymore for installing brew, we should separate them from the default so the file can be deleted on downstream images.

Please [read the Contributor's Guide](https://universal-blue.org/CONTRIBUTING/) before submitting a pull request.


m2Giles:
This isn't entirely necessary now. However, we should still consider splitting these off.

castrojo (approved):

EyeCantCU (approved):

PR #287, opened 2024-06-07
feat: add support for homebrew command not found

piperswe:
related to ublue-os/bluefin#1327

### DIFF
--- a/build/ublue-os-just/05-brew.just
+++ b/build/ublue-os-just/05-brew.just
@@ -28,3 +28,32 @@ install-brew:
 remove-brew:
     echo "Removing homebrew ..."
     /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh)"
+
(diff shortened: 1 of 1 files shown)


castrojo:
Thanks! Hey I think we want this to just be on the image right? The `ujust brew` command doesn't exist anymore. cc @m2Giles

m2Giles:
It still exists but isn't really needed.

We can bake something like this into bluefin

tulilirockz:
From what I know the `ujust brew` command does not exist here anymore, right? Maybe this issue can be closed

HikariKnight:
> From what I know the `ujust brew` command does not exist here anymore, right?

correct, the remove brew command is a leftover to just cover for people who had the pre-integration install.

bsherman:
Looks like this was already added at some point... 
It exists here now: https://github.com/ublue-os/packages/blob/main/packages/ublue-recipes/src/recipes/apps.just#L22-L69

PR #290, opened 2024-06-22
feat(just): add flatpak list to device-info

castrojo:
This will make it more clear when someone has done an unsupported rebase and is missing the right flatpaks.


fiftydinar:
Maybe add the information on which type of remote (user or system) & which name of the remote is used for flatpaks

bsherman (approved):

KyleGospo (approved):

PR #296, opened 2024-07-02
chore: update cosign public key

bsherman:
Relates: ublue-os/main#599

This is the FIRST of many PRs which need to merged related to the linked issue.

As of my initial description, we MAY still need modifications to the policy file to support both the old and new pub keys in containers policy. Thank you @KyleGospo for calling this out.

@p5 may be implement this. or I.


EyeCantCU (dismissed):

m2Giles (approved):

files/usr/etc/pki/containers/ublue-os.pub
p5, line 1: I believe we want to move this to another name, and keep the old public key here.

We then need to figure out how to push an update to policy.json to change `keyPath` to `keyPaths` array

files/usr/etc/containers/policy.json
m2Giles, line 1: This is giving me an error with skopeo.

image

EyeCantCU (approved):

m2Giles (approved):

PR #298, opened 2024-07-02
chore: add fix-key-and-update.sh script to resolve cosign pub key issues

p5:

Adds a script which can be executed by users who are on an image signed by an older cosign key pair.

### DIFF
--- a/scripts/fix-key-and-update.sh
+++ b/scripts/fix-key-and-update.sh
@@ -24,6 +24,3 @@ sed -i.bak "s#/usr/etc/pki/containers/ublue-os.pub#/etc/pki/containers/ublue-os.
 # Update system, respecting new public signing key.
 echo "Updating the system..."
 rpm-ostree update
-
-# Clean up
-rm /etc/containers/policy.json.bak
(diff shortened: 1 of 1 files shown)


p5:
Doing one final pass in a VM, so please do not merge yet :)

scripts/fix-key-and-update.sh
```diff
@@ -0,0 +1,29 @@
...
+# Ensure the path to the public key matches the local copy location.
+echo "Updating the path to the public key in the container policy..."
+sed -i.bak "s#/usr/etc/pki/containers/ublue-os.pub#/etc/pki/containers/ublue-os.pub#" /etc/containers/policy.json
+
+# Update system, respecting new public signing key.
+echo "Updating the system..."
+rpm-ostree update
+
+# Clean up
+rm /etc/containers/policy.json.bak
```
bsherman, line 29: Is it necessary to remove the backup file? If we are concerned about it being too generic, perhaps use a different backup filename, but seems like leaving it in place provides a means to rollback should that HAVE to happen?

p5: Good point.  Leaving the backup file in place
[7dde7e0](https://github.com/ublue-os/config/pull/298/commits/7dde7e0b89f214f29b1d3e43eb95e1539734c48a)

KyleGospo (approved):

m2Giles (approved):

PR #302, opened 2024-07-04
refactor: Rewrite luks tpm2 script.

m2Giles:
The previous version of the tpm2 script did not fail out when encountering unsupported configurations. This script is narrowly focused at enabling tpm2 decryption for root devices. By default we expect this to be configured via the installer.

This script will now fail out if it cannot find the disk. The disk is determined via the kernel commandline instead of /etc/crypttab. It still only supports a single disk. It now also provides hints on why the script has failed and the commands to run to disable and/or enable.

Please [read the Contributor's Guide](https://universal-blue.org/CONTRIBUTING/) before submitting a pull request.

### DIFF
--- a/build/ublue-os-luks/luks-disable-tpm2-autounlock
+++ b/build/ublue-os-luks/luks-disable-tpm2-autounlock
@@ -31,7 +31,14 @@ if ! grep -q "${RD_LUKS_UUID}" <<< "$(lsblk)" ; then
 fi
 
 # Cut off the luks-
-DISK_UUID=${RD_LUKS_UUID:5}
+LUKS_PREFIX="luks-"
+if grep -q ^${LUKS_PREFIX} <<< "${RD_LUKS_UUID}"; then
+  DISK_UUID=${RD_LUKS_UUID#"$LUKS_PREFIX"}
+else
+  echo "LUKS UUID format mismatch."
+  echo "Exiting..."
+  exit 1
+fi
 
 # Specify Crypt Disk by-uuid
 CRYPT_DISK="/dev/disk/by-uuid/$DISK_UUID"
@@ -40,6 +47,15 @@ CRYPT_DISK="/dev/disk/by-uuid/$DISK_UUID"
 if [[ ! -L "$CRYPT_DISK" ]]; then
   printf "LUKS device not listed in block devices.\n"
   printf "Exiting...\n"
+  exit 1
+fi
+
+echo "Ensuring you have a known passphrase in key-slot 0"
+if ! cryptsetup luksOpen --test-passphrase --key-slot 0 "$CRYPT_DISK"; then
+  echo "You do not have a known passphrase in key-slot 0."
+  echo "You have likely modified your luks header."
+  echo "You will have to manually use systemd-cryptenroll."
+  echo "Exiting..."
   exit 1
 fi
 
--- a/build/ublue-os-luks/luks-disable-tpm2-autounlock
+++ b/build/ublue-os-luks/luks-disable-tpm2-autounlock
@@ -1,17 +1,45 @@
 #!/bin/bash
 ## disable auto-unlock LUKS2 encrypted root on Fedora/Silverblue/maybe others
-set -u
+set -euo pipefail
 
 [ "$UID" -eq 0 ] || { echo "This script must be run as root."; exit 1;}
 
 read -p "This will modify your system and disable TPM2 auto-unlock of your LUKS partition! Are you sure you've read the script and are good with this? " -n 1 -r
 echo
 if [[ ! $REPLY =~ ^[Yy]$ ]]; then
-  [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1 # handle exits from shell or function but don't exit interactive shell
+  [[ "$0" = "${BASH_SOURCE[0]}" ]] && exit 1 || return 1 # handle exits from shell or function but don't exit interactive shell
 fi
 
-DISK_UUID=$(sudo awk '{ print $2 }' /etc/crypttab | cut -d= -f2)
+## Inspect Kernel Cmdline for rd.luks.uuid
+RD_LUKS_UUID="$(xargs -n1 -a /proc/cmdline | grep rd.luks.uuid | cut -d = -f 2)"
+
+# Check to make sure cmdline rd.luks.uuid exists
+if [[ -z ${RD_LUKS_UUID:-} ]]; then
+  printf "LUKS device not defined on Kernel Commandline.\n"
+  printf "This is not supported by this script.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
+# Check to make sure that the specified cmdline uuid exists.
+if ! grep -q "${RD_LUKS_UUID}" <<< "$(lsblk)" ; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
+# Cut off the luks-
+DISK_UUID=${RD_LUKS_UUID:5}
+
+# Specify Crypt Disk by-uuid
 CRYPT_DISK="/dev/disk/by-uuid/$DISK_UUID"
+
+# Check to make sure crypt disk exists
+if [[ ! -L "$CRYPT_DISK" ]]; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
 
 ## Restore the crypttab
 cp -a /etc/crypttab /etc/crypttab.working-before-disable-tpm2
@@ -23,17 +51,15 @@ else
 fi
 
 ## Wipe luks slot
-cryptsetup luksDump $CRYPT_DISK | grep systemd-tpm2 > /dev/null
-if [ 0 -eq $? ]; then
+if cryptsetup luksDump "$CRYPT_DISK" | grep systemd-tpm2 > /dev/null; then
   echo "Wiping systemd-tpm2 from LUKS on $CRYPT_DISK"
-  systemd-cryptenroll --wipe-slot=tpm2 $CRYPT_DISK
+  systemd-cryptenroll --wipe-slot=tpm2 "$CRYPT_DISK"
 else
   echo "No systemd-tpm2 found in LUKS to wipe"
 fi
 
 ## Disable initramfs
-rpm-ostree initramfs | grep tpm2 > /dev/null
-if [ 0 -eq $? ]; then
+if rpm-ostree initramfs | grep tpm2 > /dev/null; then
   echo "WARNING: if you configured initramfs for anything other than TPM2, this wipes that too..."
   echo "here's a printout:"
   rpm-ostree initramfs
@@ -41,5 +67,8 @@ if [ 0 -eq $? ]; then
   echo "Disabling rpm-ostree initramfs..."
   rpm-ostree initramfs --disable
 else
-  echo "TPM2 was not found in 'rpm-ostree initramfs'..."
+  echo "TPM2 is not configured in 'rpm-ostree initramfs'..."
 fi
+
+echo "TPM2 auto-unlock disabled..."
+echo "To reenroll TPM2 auto unlock please run ujust setup-luks-tpm-unlock..."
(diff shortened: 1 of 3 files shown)


m2Giles:
This only supports single disk decryption.

If we want to support multi-disk decryption. We can instead parse the output of lsblk.

For example
`lsblk -o NAME,TYPE -n -i -r | tr ' ' '='`

build/ublue-os-luks/luks-enable-tpm2-autounlock
```diff
@@ -1,57 +1,65 @@
 #!/bin/bash
 ## setup auto-unlock LUKS2 encrypted root on Fedora/Silverblue/maybe others
-set -u
+set -eou pipefail
 
 [ "$UID" -eq 0 ] || { echo "This script must be run as root."; exit 1;}
 
 read -p "This will modify your system and enable TPM2 auto-unlock of your LUKS partition! Are you sure you've read the script and are good with this? " -n 1 -r
```
bsherman, line 7: Should we remove "you sure you've read the script and" and instead state that we only use PCR7 which and give some idea what will cause the auto-unlock to fail and require a manual unlock?

bsherman (changes requested):

bsherman:
> This only supports single disk decryption.
> 
> If we want to support multi-disk decryption. We can instead parse the output of lsblk.
> 
> For example `lsblk -o NAME,TYPE -n -i -r | tr ' ' '='`

I really like the current changes... much better parsing the cmdline vs the crypttab.

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,18 +1,46 @@
...
+# Specify Crypt Disk by-uuid
 CRYPT_DISK="/dev/disk/by-uuid/$DISK_UUID"
 
+# Check to make sure crypt disk exists
+if [[ ! -L "$CRYPT_DISK" ]]; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
```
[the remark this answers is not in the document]
m2Giles: Yeah I think removing that echo for the restore makes sense.

build/ublue-os-luks/luks-enable-tpm2-autounlock
m2Giles: We should indicate that this will bind to secureboot state via PCR7. That will give a user some idea on what this is doing.

And probably specify that this is using systemd-cryptenroll so a user can look up information about it.

Also, clearly users are not inspecting the script prior to running so I'm unsure if that warning is actually serving any purpose right now.

build/ublue-os-luks/luks-enable-tpm2-autounlock
```diff
@@ -1,57 +1,70 @@
...
 fi
-cp -a /etc/crypttab /etc/crypttab.known-good
 
-## modify the crypttab
-grep tpm2-device /etc/crypttab > /dev/null
-if [ 0 -eq $? ]; then
-  echo "TPM2 already present in /etc/crypttab. Exiting..."
-  [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+# Cut off the luks-
+DISK_UUID=${RD_LUKS_UUID:5}
```
qoijjj, line 37: It might be good to validate that the RD_LUKS_UUID starts with "luks-" in case this format ever changes in the future

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,45 +1,74 @@
...
+# Check to make sure that the specified cmdline uuid exists.
+if ! grep -q "${RD_LUKS_UUID}" <<< "$(lsblk)" ; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
+# Cut off the luks-
+DISK_UUID=${RD_LUKS_UUID:5}
```
qoijjj, line 34: same thing here

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,45 +1,74 @@
...
   echo "Restoring /etc/crypttab.known-good to original /etc/crypttab"
   mv /etc/crypttab.known-good /etc/crypttab
-else
-  echo "No /etc/crypttab.known-good found to restore"
 fi
 
 ## Wipe luks slot
-cryptsetup luksDump $CRYPT_DISK | grep systemd-tpm2 > /dev/null
-if [ 0 -eq $? ]; then
+if cryptsetup luksDump "$CRYPT_DISK" | grep systemd-tpm2 > /dev/null; then
```
qoijjj, line 54: This validation can be more thorough. For example, if the user accidentally enrolled multiple tokens, or has no backup passphrase.

In the latter case, I believe systemd-cryptenroll will throw an error, which would break the script if not handled

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,45 +1,74 @@
...
 
+# Check to make sure crypt disk exists
+if [[ ! -L "$CRYPT_DISK" ]]; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
 ## Restore the crypttab
 cp -a /etc/crypttab /etc/crypttab.working-before-disable-tpm2
```
qoijjj, line 47: can this be deleted?

build/ublue-os-luks/luks-enable-tpm2-autounlock
qoijjj:
```
LUKS_PREFIX="luks-"
DISK_UUID="${RD_LUKS_UUID#$LUKS_PREFIX}"
```

something like this might be more elegant

build/ublue-os-luks/luks-disable-tpm2-autounlock
qoijjj:
```


DISK_LUKSDUMP=$(cryptsetup luksDump $CRYPT_DISK)

if echo $DISK_LUKSDUMP | grep fido > /dev/null || echo $DISK_LUKSDUMP | {add logic to check for passphrase here}; then
  echo "cryptsetup has been manually configured, exiting..."
  exit 1
fi
```

m2Giles: This is the same logic that we had previously. We can put a warning at the top to make sure you still have a decryption method besides the tpm.

We could have a warning during the enrollment to always ensure that you have a password or recovery key to ensure you have access to the data.

build/ublue-os-luks/luks-enable-tpm2-autounlock
m2Giles: I like this.

build/ublue-os-luks/luks-disable-tpm2-autounlock
m2Giles: For those who have the crypttab restored. It's probably good to have a backup of the crypttab before this was run

build/ublue-os-luks/luks-disable-tpm2-autounlock
qoijjj: @m2Giles I don't see this logic in the previous script, unless you mean the backing up of the previous crypttab?

build/ublue-os-luks/luks-enable-tpm2-autounlock
qoijjj: sanity checks never hurt :smile: especially when messing with cryptsetup

build/ublue-os-luks/luks-disable-tpm2-autounlock
m2Giles: I do like the check.

I think the simple thing will be checking if passphrase is in slot 0. If not it's safe to assume the user manually configured the device.

qoijjj: @m2Giles yeah that's a reasonable assumption.

qoijjj: @m2Giles I think this script should be as foolproof as possible, so if there's *any* sign that the user has manually touched cryptsetup, we should fail and say "it looks like you've got this covered :)"

otherwise you open this up to needing to support custom configurations

qoijjj (approved):

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,45 +1,90 @@
...
 CRYPT_DISK="/dev/disk/by-uuid/$DISK_UUID"
 
+# Check to make sure crypt disk exists
+if [[ ! -L "$CRYPT_DISK" ]]; then
+  printf "LUKS device not listed in block devices.\n"
+  printf "Exiting...\n"
+  exit 1
+fi
+
+echo "Ensuring you have a known passphrase in key-slot 0"
```
bsherman, line 53: This check seems redundant as `systemd-cryptenroll` requires the user to provide a valid description method (passphrase, recovery key, fido, etc) before it will add TPM2.

build/ublue-os-luks/luks-enable-tpm2-autounlock
```diff
@@ -1,57 +1,85 @@
...
+  exit 1
 fi
-cp -a /etc/crypttab /etc/crypttab.known-good
 
-## modify the crypttab
-grep tpm2-device /etc/crypttab > /dev/null
-if [ 0 -eq $? ]; then
-  echo "TPM2 already present in /etc/crypttab. Exiting..."
-  [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+echo "Ensuring you have a known passphrase in key-slot 0."
```
bsherman, line 56: This check seems redundant as systemd-cryptenroll requires the user to provide a valid description method (passphrase, recovery key, fido, etc) before it will add TPM2.

build/ublue-os-luks/luks-disable-tpm2-autounlock
```diff
@@ -1,45 +1,90 @@
...
-  echo "No /etc/crypttab.known-good found to restore"
 fi
 
 ## Wipe luks slot
-cryptsetup luksDump $CRYPT_DISK | grep systemd-tpm2 > /dev/null
-if [ 0 -eq $? ]; then
+if cryptsetup luksDump "$CRYPT_DISK" | grep systemd-tpm2 > /dev/null; then
   echo "Wiping systemd-tpm2 from LUKS on $CRYPT_DISK"
-  systemd-cryptenroll --wipe-slot=tpm2 $CRYPT_DISK
+  systemd-cryptenroll --wipe-slot=tpm2 "$CRYPT_DISK"
```
bsherman, line 72: Confirming a comment I made elsewhere, `systemd-cryptenroll` will NOT wipe the tpm2 slot if it's the last remaining slot.

```
$ systemd-cryptenroll /dev/disk/by-uuid/d4f1e47c-a653-4cd5-a6ae-bfc62c0a6852 --wipe-slot=tpm2 
Wipe operation would leave no valid slots around, can't allow that, sorry.
```

Granted, if a user is living dangerously if they get to this state, and I think it would have to be by use of having a tpm2 slot and recover-key slot both setup by systemd-cryptenroll, and then having wiped the recovery slot.

Similarly, `cryptsetup luksKillSlot` prevents removing of a slot using that slot's passphrase.

bsherman: As an aside, if a user does reach the state of only a tpm2 slot... a passphrase can be re-added:
`cryptsetup luksAddKey --token-id 2 --token-type systemd-tpm2 /dev/disk/by-uuid/d4f1e47c-a653-4cd5-a6ae-bfc62c0a6852`

build/ublue-os-luks/luks-enable-tpm2-autounlock
bsherman: Actually this is worse than redundant as it force the user to only have a passphrase in slot 0... So if a user added a recovery key and removed the original passphrase, this fails, but i think our script shouldn't fail for that use case.

bsherman (changes requested):
https://github.com/ublue-os/config/pull/302/files#r1667429020

Please remove pre-emptive test for passphrase in key-slot 0... as systemd-cryptenroll handles this adequately.

bsherman (approved):
LGTM

KyleGospo (approved):

PR #310, opened 2024-08-07
fix: Drop use of /usr/etc/ and instead place files in /etc

KyleGospo:
ostree handles the conversion to /usr/etc automatically


castrojo (dismissed):

scripts/fix-key-and-update.sh
```diff
@@ -19,7 +19,7 @@ curl https://raw.githubusercontent.com/ublue-os/main/main/cosign.pub > /etc/pki/
 
 # Ensure the path to the public key matches the local copy location.
 echo "Updating the path to the public key in the container policy..."
-sed -i.bak "s#/usr/etc/pki/containers/ublue-os.pub#/etc/pki/containers/ublue-os.pub#" /etc/containers/policy.json
+sed -i.bak "s#/etc/pki/containers/ublue-os.pub#/etc/pki/containers/ublue-os.pub#" /etc/containers/policy.json
```
p5, line 22: I think this is still needed.  It's pattern matching on the old string (containing /usr)

KyleGospo: Fixed, thanks

p5 (approved):

m2Giles (approved):

PR #311, opened 2024-08-08
fix: Move policy.json back to /usr/etc temporarily while issues are worked out

KyleGospo:
Without this, this package conflicts with containers-common

We should likely repackage that with our changes and have this package provide it to fix this properly.


p5 (dismissed):

p5 (approved):

cgwalters:
xref https://github.com/containers/bootc/issues/942#issuecomment-2525225513 

More generally a few things here. One core problem is that `policy.json` would really be nicer as a `.d` file with drop-ins, and that would have simplified a lot here.

Or even just having a `/usr` and `/etc` split and moving the defaults into `/usr`; an issue exists for that in https://github.com/containers/image/issues/2157

Anyways, my short term recommendation here is just don't ship this config as an RPM; instead use the "artifact pattern" for it and copy it over or so, right?

fiftydinar:
> Anyways, my short term recommendation here is just don't ship this config as an RPM; instead use the "artifact pattern" for it and copy it over or so, right?

I confirm that just `cp`-ing `policy.json` to it's location works.

PR #328, opened 2024-10-03
docs: `ujust enroll-secure-boot-key`

Malix-Labs:
General enhancement and precision about

> *QWERTY* keyboard input and navigation

Also see https://github.com/ublue-os/bluefin-docs/pull/28

### DIFF
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -32,7 +32,7 @@ regenerate-grub:
 # Enroll Nvidia driver & KMOD signing key for secure boot - Enter password "universalblue" if prompted
 enroll-secure-boot-key:
     sudo mokutil --timeout -1
-    echo 'Enter password "universalblue" if prompted next'
+    echo 'The next line will prompt for a MOK password. Then, input "universalblue"'
     sudo mokutil --import /etc/pki/akmods/certs/akmods-ublue.der
     echo 'At next reboot, the mokutil UEFI menu UI will be displayed (*QWERTY* keyboard input and navigation).\nThen, select "Enroll MOK", and input "universalblue" as the password'
 
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -31,11 +31,10 @@ regenerate-grub:
 
 # Enroll Nvidia driver & KMOD signing key for secure boot - Enter password "universalblue" if prompted
 enroll-secure-boot-key:
-    echo 'Enter password "universalblue" if prompted after your user password.'
     sudo mokutil --timeout -1
+    echo 'Enter password "universalblue" if prompted next'
     sudo mokutil --import /etc/pki/akmods/certs/akmods-ublue.der
-    echo 'When you reboot your computer, follow the instructions to start MOK util'
-    echo 'by pressing a key, then enroll the secure boot key and enter "universalblue" as the password'
+    echo 'At next reboot, the mokutil UEFI menu UI will be displayed (*QWERTY* keyboard input and navigation).\nThen, select "Enroll MOK", and input "universalblue" as the password'
 
 # Install or uninstall distrobox-git
 setup-distrobox-git:


build/ublue-os-just/00-default.just
```diff
@@ -31,11 +31,10 @@ regenerate-grub:
 
 # Enroll Nvidia driver & KMOD signing key for secure boot - Enter password "universalblue" if prompted
 enroll-secure-boot-key:
-    echo 'Enter password "universalblue" if prompted after your user password.'
     sudo mokutil --timeout -1
+    echo 'Enter password "universalblue" if prompted next'
```
Malix-Labs, line 35: > after your user password

Is not needed anymore, since it was already prompted in the previous line

Malix-Labs, line 35: Would [L36](https://github.com/Malix-Labs/config/blob/30f286b34eb93173ef4ba4581eefc7a73586a0e6/build/ublue-os-just/00-default.just#L36) always prompt for the MOK password in all cases ?
If yes, then I also suggest this :
```suggestion
    echo 'The next line will prompt for a MOK password. Then, input "universalblue"'
```

bsherman: Your suggestion is good, as yes, `mokutil  --import` will always prompt for a password.

build/ublue-os-just/00-default.just
```diff
@@ -31,11 +31,10 @@ regenerate-grub:
...
     sudo mokutil --import /etc/pki/akmods/certs/akmods-ublue.der
-    echo 'When you reboot your computer, follow the instructions to start MOK util'
-    echo 'by pressing a key, then enroll the secure boot key and enter "universalblue" as the password'
+    echo 'At next reboot, the mokutil UEFI menu UI will be displayed (*QWERTY* keyboard input and navigation).\nThen, select "Enroll MOK", and input "universalblue" as the password'
```
bsherman, line 37: This is a good improvement. However, the ` (*QWERTY* keyboard input and navigation)` section may not be clear and may be too much detail. Please remove that section. Thanks!

bsherman (changes requested):

build/ublue-os-just/00-default.just
Malix-Labs: It was the reason I originally opened that PR actually

In my zone, AZERTY is the standard, and it took me hours to understand why the password were not matching - ARERTY("universalblue") == QWERTY("universqlblue")

It seems to have bitten other people too;

See https://universal-blue.discourse.group/t/mokutil-list-enrolled-explanation/4266/18?u=malix

How do you suggest this information should be included better ?

bsherman: Hmm... apologies, as I'd missed the detail around that topic.

The MOK import is problematic for a number of reasons. This is another one added to the list. :-(

If the script always set the MOK password to "universalblue" (QWERTY)  does that help? 

I assume it would not, as likely the problem is the UEFI/MOK firmware not respecting keyboard layouts?

Malix-Labs: > the problem is the UEFI/MOK firmware not respecting keyboard layouts?

The problem indeed comes from the mokutil UEFI menu UI not respecting keyboard layouts (always QWERTY)

So for example, my AZERY "a" is translated as a QWERTY "q" there (so "universqlblue")

And as the password input UI doesn't let you see character, it was impossible to know before [someone on the forum told me it could come from there](https://universal-blue.discourse.group/t/mokutil-list-enrolled-explanation/4266/18?u=malix), after having wasted 15 minutes trying with different methods :sweat_smile:

HikariKnight: i wonder if setting the password to `ublue` will be good enough to avoid issues with AZERTY and QWERTZ layouts

bsherman: @HikariKnight 

That would work but `ublue` is too short. I think the minimum is 8 chars?

bsherman (approved):
I didn't mean to let this sit for past weeks. 

There's not an easy solution, and upstream Fedora has a similar issue where they call out in the installer that early boot is QUERTY only.

Malix-Labs:
Indeed, a idempotent password should be thought about in a different issue.
Should I create it ?

bsherman:
> Indeed, an idempotent password should be thought about in a different issue. Should I create it ?

Sure, it doesn't hurt to have a place holder for discussion.

Malix-Labs:
See https://github.com/ublue-os/config/issues/337

HikariKnight (approved):

PR #333, opened 2024-10-30
feat(just): Use `run0` instead of `sudo` in scripts

fiftydinar:
It is more secure than `sudo` & it uses native password dialog.

`run0` currently always asks for the password per root-call at the moment, so I modified the scripts which used `sudo` to initiate 1 call only. So user will only see 1 password prompt. This may make some scripts slightly less readable, but it's still easy to read imo.

I added `librunasroot`, which only uses `run0` if it's installed, while fallbacking to `sudo` if not. This makes sure that <=F40 images won't be affected badly with this change.

RPM is updated to reflect this change.

### DIFF
--- a/build/ublue-os-just/lib-ujust/librun0.sh
+++ b/build/ublue-os-just/lib-ujust/librun0.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/bash
+
+# Function to use run0 instead of sudo, if installed (since F41)
+# run0 is more secure than using sudo, so it's recommend to use it instead of sudo
+# The current caveat is that run0 always requires password authentication per root-call (--no-ask-password doesn't work at the moment)
+# so use it only for one-liner commands (bash -c can help here, with potential of less readable code)
+# See:
+# https://github.com/systemd/systemd/issues/33366
+# https://github.com/polkit-org/polkit/issues/472
+
+run0() {
+if command -v run0 &> /dev/null; then
+  run0 "${@}"
+else
+  sudo "${@}"
+fi  
+}
--- a/build/ublue-os-just/lib-ujust/librun0.sh
+++ b/build/ublue-os-just/lib-ujust/librun0.sh
@@ -10,7 +10,7 @@
 
 run0() {
 if command -v run0 &> /dev/null; then
-  run0 "${@}"
+  run0 --background="" "${@}"
 else
   sudo "${@}"
 fi  
--- a/build/ublue-os-just/lib-ujust/ujust.sh
+++ b/build/ublue-os-just/lib-ujust/ujust.sh
@@ -10,3 +10,5 @@ source /usr/lib/ujust/libfunctions.sh
 source /usr/lib/ujust/libdistrobox.sh
 # Import functionality related to toolbox
 source /usr/lib/ujust/libtoolbox.sh
+# Import functionality related to usage of run0
+source /usr/lib/ujust/librun0.sh
--- a/build/ublue-os-just/lib-ujust/ujust.sh
+++ b/build/ublue-os-just/lib-ujust/ujust.sh
@@ -11,4 +11,4 @@ source /usr/lib/ujust/libdistrobox.sh
 # Import functionality related to toolbox
 source /usr/lib/ujust/libtoolbox.sh
 # Import functionality related to usage of run0
-source /usr/lib/ujust/librun0.sh
+source /usr/lib/ujust/librunasroot.sh
--- a/build/ublue-os-just/00-default.just
+++ b/build/ublue-os-just/00-default.just
@@ -14,26 +14,25 @@ bios:
 
 # Show all messages from this boot
 logs-this-boot:
-    sudo journalctl -b 0
+    run0 journalctl -b 0
 
 # Show all messages from last boot
 logs-last-boot:
-    sudo journalctl -b -1
+    run0 journalctl -b -1
 
 # Regenerate GRUB config, useful in dual-boot scenarios where a second operating system isn't listed
 regenerate-grub:
     #!/usr/bin/bash
     if [ -d /sys/firmware/efi ]; then
-      sudo grub2-mkconfig -o /etc/grub2-efi.cfg
+      run0 grub2-mkconfig -o /etc/grub2-efi.cfg
     else
-      sudo grub2-mkconfig -o /etc/grub2.cfg
+      run0 grub2-mkconfig -o /etc/grub2.cfg
     fi
 
 # Enroll Nvidia driver & KMOD signing key for secure boot - Enter password "universalblue" if prompted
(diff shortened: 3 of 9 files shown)


fiftydinar:
Note: `run0` colors the output in red by default, to indicate the it's ran as root.

This can be easily disabled with `--background=""` flag if desired

build/ublue-os-just/lib-ujust/librun0.sh
```diff
@@ -0,0 +1,17 @@
...
+# run0 is more secure than using sudo, so it's recommend to use it instead of sudo
+# The current caveat is that run0 always requires password authentication per root-call (--no-ask-password doesn't work at the moment)
+# so use it only for one-liner commands (bash -c can help here, with potential of less readable code)
+# See:
+# https://github.com/systemd/systemd/issues/33366
+# https://github.com/polkit-org/polkit/issues/472
+
+run0() {
+if command -v run0 &> /dev/null; then
+  run0 --background="" "${@}"
```
qoijjj, line 13: what does --background do? don't we want stdout?

build/ublue-os-just/lib-ujust/librun0.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Function to use run0 instead of sudo, if installed (since F41)
+# run0 is more secure than using sudo, so it's recommend to use it instead of sudo
+# The current caveat is that run0 always requires password authentication per root-call (--no-ask-password doesn't work at the moment)
+# so use it only for one-liner commands (bash -c can help here, with potential of less readable code)
+# See:
+# https://github.com/systemd/systemd/issues/33366
+# https://github.com/polkit-org/polkit/issues/472
+
+run0() {
```
qoijjj, line 11: maybe a slightly different name just to avoid clashes?

build/ublue-os-just/lib-ujust/librun0.sh
fiftydinar: It shows the output without color, unlike the default, which is red.
That's what Lennart recommended to use.

build/ublue-os-just/lib-ujust/librun0.sh
fiftydinar: I can rename the function to `run-as-root`, along with the library rename `librunasroot.sh`?

Is that good?

fiftydinar: Just renamed it to `run-as-root`

fiftydinar:
Another regression with `run0` unrelated to this PR (where everything I tested works):

`run0 dmidecode` doesn't output anything, while `sudo dmidecode` does.

So @antheas is indeed right that there can be some regressions with `run0` & that's why I'm proposing a small transition like this, which wouldn't affect anyone badly.

antheas:
run0 has other differences. It does not inherit your environment (e.g., there is no `sudo -E`) and there is no direct pipe to the executable as it uses IPC

It is more similar to running a command over ssh than sudo.

I think the intent is for novice users to perform routine maintenance on their computers and not so much a replacement for sudo

castrojo:
Yeah I fail to understand what problem we're trying to solve here?

fiftydinar:
@castrojo
I'll let @RoyalOughtness explain the security benefits of using `run0` instead of `sudo` once he's available, as he's more knowledgable in this topic.

It's not a problem to solve per-see, but it's to improve the security of just scripts. That's all.

bsherman:
> @castrojo I'll let @RoyalOughtness explain the security benefits of using `run0` instead of `sudo` once he's available, as he's more knowledgable in this topic.
> 
> It's not a problem to solve per-se, but it's to improve the security of just scripts. That's all.

@fiftydinar I appreciate you and @RoyalOughtness working to improve security... but the reason secureblue is a distinct project is that it makes changes in the name of security which also limit usability.

In general, adding more conditionals to a code base makes it harder to maintain.

If `run0` becomes standard in upstream Fedora or the ublue team agrees we want `run0` in main images, we could possibly revisit this, but adding this seems now like a bad idea to me.

fiftydinar:
@bsherman Understandable (especially the part about adding more conditionals).

If most people agree to close this (I see 3 people here disagreeing with this change at the moment), then let's do that.

RoyalOughtness:
run0 isn't a drop-in sudo replacement, so I agree this should be closed. 

>  ublue team agrees we want run0 in main images

Also, run0 is already in your images :P

build/ublue-os-just/lib-ujust/ujust.sh
```diff
@@ -10,3 +10,5 @@ source /usr/lib/ujust/libfunctions.sh
 source /usr/lib/ujust/libdistrobox.sh
 # Import functionality related to toolbox
 source /usr/lib/ujust/libtoolbox.sh
+# Import functionality related to usage of run0
+source /usr/lib/ujust/librun0.sh
```
bsherman, line 14: I don't see `/usr/lib/ujust/librun0.sh` in existing images or even in this PR. Not sure how this was going to work, unless it was what got renamed to `librunasroot.sh`.

fiftydinar: Yes, I forgot to rename it, thanks for noticing

fiftydinar: Fixed

bsherman:
> Also, run0 is already in your images :P

@RoyalOughtness ah, it's on F41 images. It's part of systemd 256.

@fiftydinar did mention "<= F40" but it didn't click that "run0 was added in systemd 256 which is in Fedora 41".  And I run on bluefin stable images except for testing so wasn't seeing this anywhere and was quite confused.

At least THAT confusion is sorted out.

> run0 isn't a drop-in sudo replacement, so I agree this should be closed.

:+1:

bsherman:
OK, given recent discussion, I think we're good to close this.

Without a burning need, we should wait to implement any systemd 256 specific changes until another release cycle when we no longer have Fedora 40 builds happening.

At that time, if it makes sense to use run0 instead of sudo (though @RoyalOughtness already suggested it wasn't a drop-in replacement) we can entertain a new PR.

RoyalOughtness:
@bsherman in case you're curious, the functionality run0 is based on has been in systemd for awhile (systemd-run), run0 is a more familiar frontend to it.

and while it isn't a drop-in replacement for sudo, it does make it feasible to run a linux system completely sudoless, eliminating the need for a suid root binary. I know there are Fedora maintainers who want to see suid-root binaries done away with in the long term, so eventually I would expect the standard to change. However, I expect it will take years to become the default, like we saw with the wayland transition

Also a related and interesting read: https://tim.siosm.fr/blog/2023/12/19/ssh-over-unix-socket/

PR #334, opened 2024-11-01
fix(just): Alias `just` to `ujust`, to avoid non-working `just` command

fiftydinar:
Useful for people who just type `just` instead of `ujust`.


fiftydinar:
Forgot about what `ujust` wrapper actually does, I thought it only enforces bash if other shell is installed, so I'll close this.
I may reopen If I find the better solution to solve this.

`/usr/bin/just --justfile /usr/share/ublue-os/justfile "${@}"`

fiftydinar:
Improved the logic to mitigate the scenario above & reopened

fiftydinar:
Just found out that `zsh` doesn't support exporting functions.
Will try to see how to mitigate this.

fiftydinar:
Closing, because it won't work with `zsh` & due to this being a hack, rather than a proper fix.

Issue #326, opened 2024-10-02
Fix ujust TPM error message


RPGFSjosh:
When running `ujust setup-luks-tpm-unlock` with a key already enrolled, it returns this error:
```
Remove the existing TPM2 enrollment with ujust remove-luks-tpm2-autounlock
```
The correct command to resolve the error has since changed to:
```
ujust remove-luks-tpm-unlock
```

PR #336, opened 2024-11-02
fix: luks allow wipe for re-enroll and improve prompt/output

bsherman:
luks tpm2 auto lock/unlock scripts should not actually reference ujust since they can be used without it.

if user has found these ujust recipes, it should be obvious what to do from the ujust list of recipes.

Also, why not wipe and re-enroll. 

Closes: #326


antheas:
Speaking of, might as well fix this instead of tweaking the message.

The script should re-enroll the signature if its called again. It's very annoying to have to call unenroll when the hash gets invalidated to re-enroll it.

antheas:
I think you can remove the whole block that checks if tpm2 is enrolled and just append `--wipe-slot=tpm2` in the place you enroll.

castrojo (approved):

m2Giles (approved):

PR #350, opened 2025-01-10
fix(just): incorrect config files in configure-broadcom-wl

jorti:
fix: When disabling the wl driver, the files have to be written in /etc/modprobe.d also use a symlink to /dev/null to override the distribution-provided config.


HikariKnight (approved):

bsherman:
@HikariKnight are we even shipping the `wl` driver in any images today?

I think Bluefin removed it.

bsherman:
Answering my own question... Aurora and Bazzite both still have it... 

Maybe the right answer is shipping the `wl` kmod, but in a pre-disabled state (what this ujust recipe would do)... allowing this recipe to enable if desired.

HikariKnight:
> Answering my own question... Aurora and Bazzite both still have it...
> 
> Maybe the right answer is shipping the `wl` kmod, but in a pre-disabled state (what this ujust recipe would do)... allowing this recipe to enable if desired.

this is what we do in bazzite i believe

bsherman:
Closed by https://github.com/ublue-os/packages/pull/229

PR #353, opened 2025-01-16
fix(just): separate just and ujust

Danar435:
Currently the justfile in the home directory is automatically being cluttered with ujust commands. This PR separates both justfiles better, while also providing a simple explanation of how to set up just commands. The original link is dead, so I replaced it with the just GitHub link.

Let me know if we should add or change any commands. `edit` works as a simple example, but maybe we could add `help` too. 

Closes ublue-os/bluefin#2108


ledif:
This file is completely gone now after https://github.com/ublue-os/config/pull/364 has been merged. We are no longer creating a justfile in users' home directories so https://github.com/ublue-os/bluefin/issues/2108 should be closed as well.

Closing this PR for now but feel free to re-open if you feel there is still something that needs to be done.

PR #356, opened 2025-02-01
fix(just): toggle-nvk now supports nvidia-open

ledif:
Fixes https://github.com/ublue-os/aurora/issues/157.

## Before
| Image Name | Rebase Target | Is Valid Target |
| --- | --- | --- |
| `aurora-dx-nvidia` | `aurora-dx` | ✅ |
| `aurora-dx` | `aurora-dx-nvidia` | ✅ |
| `aurora-dx-nvidia-open` | `aurora-dx-open` | ❌ |

## After
| Image Name | Rebase Target | Is Valid Target |
| --- | --- | --- |
| `aurora-dx-nvidia` | `aurora-dx` | ✅ |
| `aurora-dx` | `aurora-dx-nvidia-open` | ✅ |
| `aurora-dx-nvidia-open` | `aurora-dx` | ✅ |


HikariKnight (dismissed):

HikariKnight:
Thanks for this, when the toggle-nvk recipe was made, nvidia-open was like a year away from existing

francoism90:
It would be great to test this out. :)

m2Giles (commented):
This does mean that it won't rebase back to non-open Nvidia drivers.

HikariKnight:
in bazzite we have `/usr/libexec/nvidia-legacy-hardware` which will exit 0 (true) or exit 1 (false) depending on the user having legacy nvidia hardware.

We were going to originally use it to automatically rebase nvidia users with newer gpus to the `nvidia-open` image, but i am unsure if kyle ever did that.

this is the script
```bash
#!/usr/bin/bash
# Returns true if legacy nvidia hardware (older than turing) using die names
# GK### (Kepler), GM### (Maxwell), GP### (Pascal), GV### (Volta)
if lspci -nn | grep -P "(VGA compatible|3D) controller .+G(K|M|P|V)\d{3}" > /dev/null; then
    exit 0
fi
exit 1
```

this way we can make the toggle fail on legacy hardware

KyleGospo (dismissed):

ledif:
> in bazzite we have `/usr/libexec/nvidia-legacy-hardware` which will exit 0 (true) or exit 1 (false) depending on the user having legacy nvidia hardware.
> 
> We were going to originally use it to automatically rebase nvidia users with newer gpus to the `nvidia-open` image, but i am unsure if kyle ever did that.
> 
> this is the script
> 
> ```shell
> #!/usr/bin/bash
> # Returns true if legacy nvidia hardware (older than turing) using die names
> # GK### (Kepler), GM### (Maxwell), GP### (Pascal), GV### (Volta)
> if lspci -nn | grep -P "(VGA compatible|3D) controller .+G(K|M|P|V)\d{3}" > /dev/null; then
>     exit 0
> fi
> exit 1
> ```
> 
> this way we can make the toggle fail on legacy hardware, also unsure if Tesla cards will count as legacy, i could not find any proper info about that when i first made the script

Thanks for the snippet. I updated the PR to include it.

inffy (approved):

HikariKnight (approved):

PR #362, opened 2025-02-09
feat: add aarch64 support

GTB3NW:
https://github.com/ublue-os/main/issues/365: add aarch64 support

This is pre-requisite work required for other downstream projects to support arm devices

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,7 +154,7 @@ jobs:
 
       # Login with podman for buildah
       - name: Log in to ghcr.io
-        uses: redhat-actions/podman-login@v1
+        uses: redhat-actions/podman-login@4934294ad0449894bcd1e9f191899d7292469603 # v1
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,25 +3,21 @@ on:
   pull_request:
   merge_group:
   schedule:
-    - cron: '15 9 * * 1'  # 9:15 UTC weekly on Monday
+    - cron: "15 9 * * 1" # 9:15 UTC weekly on Monday
   workflow_dispatch:
 env:
-    IMAGE_NAME: config
-    IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
+  IMAGE_NAME: config
+  IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
 
 jobs:
-  push-ghcr:
-    name: Build and push image
-    runs-on: ubuntu-24.04
-    permissions:
-      contents: read
-      packages: write
-      id-token: write
+  calculate-tags:
+    name: Calculate tags
+    runs-on: ubuntu-latest
+    outputs:
+      alias_tags: ${{ steps.generate-tags.outputs.alias_tags }}
+      sha_short: ${{ steps.generate-tags.outputs.sha_short }}
+      date: ${{ steps.generate-tags.outputs.date }}
     steps:
-      # Checkout push-to-registry action GitHub repository
-      - name: Checkout Push to Registry action
-        uses: actions/checkout@v4
-
       - name: Generate tags
         id: generate-tags
         shell: bash
@@ -36,11 +32,53 @@ jobs:
             echo "date=$(date +%Y%m%d)" >> $GITHUB_OUTPUT
             alias_tags+=("latest")
           fi
+
           echo "alias_tags=${alias_tags[*]}" >> $GITHUB_OUTPUT
+
+  push-ghcr:
+    name: Build and push image
+    runs-on: ${{ matrix.build_arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    needs:
+      - calculate-tags
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    strategy:
+      fail-fast: false
+      matrix:
+        build_arch: [aarch64, x86_64]
+    steps:
+      # Checkout push-to-registry action GitHub repository
+      - name: Checkout Push to Registry action
+        uses: actions/checkout@v4
+
+      # Update to tags to add the arch to the tags
+      - name: Update tags
+        id: updated-tags
+        run: |
+          tags=(${{ needs.calculate-tags.outputs.alias_tags }})
+          updated_tags=()
+          for tag in "${tags[@]}"; do
+            updated_tags+=("${tag}-${{ matrix.build_arch }}")
+          done
+          echo "alias_tags=${updated_tags[*]}" >> $GITHUB_OUTPUT
+
+          echo "sha_short=${{ needs.calculate-tags.outputs.sha_short }}-${{ matrix.build_arch }}" >> $GITHUB_OUTPUT
+
+          # If date is not empty string, append the arch to the date
+          if [[ -n "${{ needs.calculate-tags.outputs.date }}" ]]; then
+            echo "date=${{ needs.calculate-tags.outputs.date }}-${{ matrix.build_arch }}" >> $GITHUB_OUTPUT
+          fi
 
       - name: Check just syntax
         id: check_just_syntax
         uses: ublue-os/just-action@v2
+
+      - name: Install skopeo & podman
+        run: |
+          sudo apt update
+          sudo apt install -y skopeo podman
 
       # Build image using Buildah action
       - name: Build Image
@@ -51,9 +89,9 @@ jobs:
             ./Containerfile
           image: ${{ env.IMAGE_NAME }}
           tags: |
-            ${{ steps.generate-tags.outputs.alias_tags }}
-            ${{ steps.generate-tags.outputs.date }}
-            ${{ steps.generate-tags.outputs.sha_short }}
+            ${{ steps.updated-tags.outputs.alias_tags }}
+            ${{ steps.updated-tags.outputs.date }}
+            ${{ steps.updated-tags.outputs.sha_short }}
           oci: true
 
       # Workaround bug where capital letters in your GitHub username make it impossible to push to GHCR.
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -3,25 +3,21 @@ on:
   pull_request:
   merge_group:
   schedule:
-    - cron: '15 9 * * 1'  # 9:15 UTC weekly on Monday
+    - cron: "15 9 * * 1" # 9:15 UTC weekly on Monday
```
GTB3NW, line 6: vs code 🤷‍♂️

.github/workflows/build.yml
```diff
@@ -3,25 +3,21 @@ on:
...
   workflow_dispatch:
 env:
-    IMAGE_NAME: config
-    IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
+  IMAGE_NAME: config
```
GTB3NW, line 9: formatting

.github/workflows/build.yml
```diff
@@ -3,25 +3,21 @@ on:
...
 
 jobs:
-  push-ghcr:
-    name: Build and push image
-    runs-on: ubuntu-24.04
-    permissions:
-      contents: read
-      packages: write
-      id-token: write
+  calculate-tags:
```
GTB3NW, line 13: opted to create a new job for this and reference the tag outputs. It was a little easier structure wise to iterate over the tags and just append the arch when needed

.github/workflows/build.yml
```diff
@@ -79,28 +117,83 @@ jobs:
...
+      # Login with docker for cosign
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
-        if: github.event_name != 'pull_request'
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      # Sign container
```
GTB3NW, line 90: moved to the manifest generation step

.github/workflows/build.yml
```diff
@@ -36,12 +32,54 @@ jobs:
...
+          # If date is not empty string, append the arch to the date
+          if [[ -n "${{ needs.calculate-tags.outputs.date }}" ]]; then
+            echo "date=${{ needs.calculate-tags.outputs.date }}-${{ matrix.build_arch }}" >> $GITHUB_OUTPUT
+          fi
+
       - name: Check just syntax
         id: check_just_syntax
         uses: ublue-os/just-action@v2
 
+      - name: Install skopeo & podman
```
GTB3NW, line 78: arm runners are nerfed a little, ran by a third party which isn't github (weird situation tbh!).

I've just opted to explicitly install both of these deps regardless of arch

.github/workflows/build.yml
```diff
@@ -79,28 +117,83 @@ jobs:
...
-        if: github.event_name != 'pull_request'
+      # Login with podman for buildah
+      - name: Log in to ghcr.io
+        uses: redhat-actions/podman-login@4934294ad0449894bcd1e9f191899d7292469603 # v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Combine manifests
```
GTB3NW, line 163: this step takes the tags with the arch appended and creates a new manifest file with both arch's, with a new tag that points to both arch versions. Client picks the appropriate one automatically

.github/workflows/build.yml
```diff
@@ -79,28 +117,83 @@ jobs:
...
+          digest=""
+
+          for tag in "${tags[@]}"; do
+            buildah manifest create ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag} \
+              ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag}-aarch64 \
+              ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag}-x86_64
+
+            buildah manifest push ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag} docker://${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag}
+
+            digest=$(skopeo inspect docker://${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}:${tag} --format '{{.Digest}}')
```
GTB3NW, line 186: fine to overwrite, all the digests will be identical

.github/workflows/build.yml
```diff
@@ -79,28 +117,83 @@ jobs:
...
+
+          echo "digest: $digest"
+
+          echo "digest=$digest" >> $GITHUB_OUTPUT
 
       - name: Sign container image
-        if: github.event_name != 'pull_request'
         run: |
-          cosign sign -y --key env://COSIGN_PRIVATE_KEY ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}@${TAGS}
+          cosign sign --recursive -y --key env://COSIGN_PRIVATE_KEY ${{ steps.registry_case.outputs.lowercase }}/${{ env.IMAGE_NAME }}@${TAGS}
```
GTB3NW, line 195: recursive flag will take the top level manifest, sign it, then find all referenced manifests and sign those too. Tested in my repo and this worked.

Screenshot 2025-02-09 at 22 30 05

GTB3NW (commented):
submitted self review to explain some decisions

bsherman:
> This is pre-requisite work required for other downstream projects to support arm devices

Hi @GTB3NW, appreciate you diving in.  Forgive me if I'm missing something obvious here, but `config` generates only noarch  artifacts not x86_64 specific, etc (at least, from my memory, I didn't double check).

I'm curious what change in this PR is actually required and why.

Aside from the formatting changes and changing out tagging is done (thank you for the explanation on those), there's also a change to signing. None of this seems specifically required for arm64.

GTB3NW:
> > This is pre-requisite work required for other downstream projects to support arm devices
> 
> Hi @GTB3NW, appreciate you diving in. Forgive me if I'm missing something obvious here, but `config` generates only noarch artifacts not x86_64 specific, etc (at least, from my memory, I didn't double check).
> 
> I'm curious what change in this PR is actually required and why.
> 
> Aside from the formatting changes and changing out tagging is done (thank you for the explanation on those), there's also a change to signing. None of this seems specifically required for arm64.

yeah sure! I don't know where you're at in terms of container knowledge so please excuse the ELI5 😆 

Containers are effectively just a zipped filesystem - which as you point out, is just noarch packages in this case, and their accompanying manifests which is what this PR looks to address. The manifests are metadata and act as pointers to the filesystems. One of those bits of metadata is the supported architectures. As far as I'm aware (and I'll comment afterwards when I check), it's not possible to have a noarch container.

Regarding signing, I _think_ the manifest is included in the signing process - IIRC I had 3 outputs when signed which were for the 3 manifests.

GTB3NW:
@bsherman the only other alternative is [this](https://github.com/redhat-actions/buildah-build/blob/main/.github/workflows/multiarch.yml#L122-L130), but it uses qemu so it super slow. I can take this approach and it should remain close to the existing pipeline, at the cost of build speed - if we don't care about that, it makes my life easier 😆

GTB3NW:
@bsherman Can confirm, noarch containers aren't possible, but that would be nice. At that point it would effectively just be a tar.gz file haha

p5:
I think the main confusion here is "why is this necessary?"

Bluefin LTS is currently making use of the ublue-os/config containers for their ARM builds, which is working because ublue-os/config doesn't contain anything architecture-specific.  
During the build, Podman looks for an ARM64 ublue-os/config image, which doesn't exists and falls back to the AMD64 one.

Personally, I feel the way forward would be to do one of two things:
1. Expose AMD64 and ARM64 under the same set of tags, so Podman intelligently pulls the correct version
2. Build only one architecture-agnostic image, using AMD64 in the metadata since it already exists, and Podman falls back to this AMD64 image during the build

I don't think exposing different arch-specific tags is necessary here since the content will be identical, except for the metadata on the image index

bsherman:
> I think the main confusion here is "why is this necessary?"
 
> Edit: I re-read the PR changes, and looks like it's already going for approach 1, which I'm okay with. But as I said, it's not strictly necessary since there's nothing arch-specific in this repo, and Podman falls back to AMD64 images when there are no ARM ones

I think @p5 caught where I'm coming from and addressed it well.

I think we're better served completing the merge of `config`/`main` repos (see: https://github.com/ublue-os/main/issues/691#issuecomment-2666759879) and potentially moving a few of the  noarch RPM builds from `config` into our `packages`/COPR pipeline... then any arch specific changes we need to make for ARM will only need to be made in two places: 1) the `akmods` repo and 2) `main`  at least, not counting COPRs and downstreams (Aurora, Bluefin, Bazzite, etc).

The changes in this PR, ensuring arch specific image metadata, will be useful in `main` and `akmods`, but I'm a -1 and making changes to this repo as we know it's going to be archived as we do the merge.

GTB3NW:
> AMD64 images when there are no ARM ones

@p5 are you sure about that? I started down this rabbit hole due to errors. I'll try find it if I can

GTB3NW:
> > I think the main confusion here is "why is this necessary?"
> 
> > Edit: I re-read the PR changes, and looks like it's already going for approach 1, which I'm okay with. But as I said, it's not strictly necessary since there's nothing arch-specific in this repo, and Podman falls back to AMD64 images when there are no ARM ones
> 
> I think @p5 caught where I'm coming from and addressed it well.
> 
> I think we're better served completing the merge of `config`/`main` repos (see: [ublue-os/main#691 (comment)](https://github.com/ublue-os/main/issues/691#issuecomment-2666759879)) and potentially moving a few of the noarch RPM builds from `config` into our `packages`/COPR pipeline... then any arch specific changes we need to make for ARM will only need to be made in two places: 1) the `akmods` repo and 2) `main` at least, not counting COPRs and downstreams (Aurora, Bluefin, Bazzite, etc).
> 
> The changes in this PR, ensuring arch specific image metadata, will be useful in `main` and `akmods`, but I'm a -1 and making changes to this repo as we know it's going to be archived as we do the merge.

@bsherman Appreciated for that context! Didn't know about the migration, I'll subscribe to that thread - if you need any help please give me a shout.

This was a sidequest for me, I just want to build an app specific OS 😆

GTB3NW:
@p5 I think you're correct (as you probably already know 😆)

https://github.com/GTB3NW/ublue-main/actions/runs/13220573956/job/36905064091#step:12:127

It was just a warning, the error I encountered was due to the shell binary. I could have sworn it errored on the FROM, but I guess not.

bsherman:
> @bsherman Appreciated for that context! Didn't know about the migration, I'll subscribe to that thread - if you need any help please give me a shout.
> 
> This was a sidequest for me, I just want to build an app specific OS 😆

ARM is a long-standing stretch goal... but thank you for appreciating our desire to clean up some tech debt before we push deeper into more build complexity.

PR #364, opened 2025-02-18
chore: remove profile.d entry that creates .justfile on $HOME

tulilirockz:
This is unnecessary because we already have ujust now and this conflicts with user-placed `Justfile`s on $HOME. Would really prefer not to touch the users $HOME directory in that way.

The documentation mentioned on that script is also outdated and may confuse some people unnecessarily


hanthor (approved):

KyleGospo (approved):

bsherman (approved):
LGTM but @tulilirockz looks like there's a conflict you'll need to resolve.

bsherman (approved):

